PR #13, opened 2022-09-27
Retrieve spectrum data from AWS

Carifio24:
This PR changes the story to pull the data for spectra from AWS, rather than using local files, and thus clears out all of the spectra from the data folder. This has the advantage of reducing the repository size. Additionally, if we decide to modify the set of galaxies at some point, no change is needed here.

To accomplish this, I've moved all of the spectrum files into an S3 bucket on AWS whose policy allows the spectrum files to be publicly viewable. In the Hubble story, the data is fetched by making a GET request to our server (which is currently just a redirect to the URL of the appropriate file in the S3 bucket, but this layer of indirection might be useful in the future). The content of the response is put into an `io.BytesIO` object, which we can then load into glue using its FITS data factory. Each spectrum file is 172 kB, so I don't think this should pose much of an issue from a network perspective.


nmearl:
This is great, @Carifio24!

PR #15, opened 2022-10-03
Remove user ability to measure spectrum wavelength until on the correct part of the story

johnarban:
Removed mouse callbacks from SpecView class and placed them into StageOne._on_marker_update. This way those callback are only active when we want them. Users can click on the spectrum once they reach marker obs_wav1


patudom:
Looking good in the spectrum viewer. We also need to disable the table viewer from listening to the click events because it is still populating measurement values even though they are not appearing in the graph. There is also a flag on the "Observed Wavelength" guideline that shouldn't let the user advance until a measurement has been made while on that particular guideline.

Screen Shot 2022-10-03 at 11 51 24 AM

johnarban:
The last commit just added to the PR fixes the issue with it adding the point to the data table, by moving the callback associated with that action to the `_on_marker_update` method. 

The flag you mentioned is still present.

PR #16, opened 2022-10-03
Remove Ell and Irr galaxy slides from Dos and Donts

johnarban:
Remove the Elliptical and Irregular galaxies from the Angular Size measurements “Do’s & Don’ts” slideshow


patudom:
Please also update the following:

- Remove the slide "Field with Multiple Objects"
- Remove the slide "Galaxy Cluster" (since almost all the galaxies we have in clusters will be elliptical)

On the "Colliding Galaxies" slide:
- Update the title to be "Field with Multiple Objects"
- Update the text to say something like "Some fields have multiple objects, such as stars or galaxies, including merging companions or galaxy cluster members."
- Update caption of the "Don't" image to say "Measure across both galaxies, and don't measure other galaxies that are not in the center in the field."

patudom:
Closed this PR because there was a merge conflict that was unresolved. See PR #20 instead.

PR #20, opened 2022-10-03
Remove ell ir gals dosdonts

johnarban:
Removes references to ellipticals and irregulars, and also updates the layout of multiple galaxies. 

Addresses the issues @patudom [mentioned](https://github.com/cosmicds/hubbleds/pull/16#issuecomment-1265708497)
> Please also update the following:
> 
> * Remove the slide "Field with Multiple Objects"
> * Remove the slide "Galaxy Cluster" (since almost all the galaxies we have in clusters will be elliptical)
> 
> On the "Colliding Galaxies" slide:
> 
> * Update the title to be "Field with Multiple Objects"
> * Update the text to say something like "Some fields have multiple objects, such as stars or galaxies, including merging companions or galaxy cluster members."
> * Update caption of the "Don't" image to say "Measure across both galaxies, and don't measure other galaxies that are not in the center in the field."


patudom:
Note, this is a redo of PR #16 to fix an unresolved merge commit.

PR #25, opened 2022-10-07
Fix bug preventing state update in selection tool callback

Carifio24:
This PR fixes a bug that could prevent `gal_selected` in the stage one state from being marked true when a galaxy is selected. There was a bug in the callback we attach to the WWT widget - if the widget's `selected_data` DataFrame is empty, then the call to `self.selected_data['name']` gives an error. This PR adds a check of whether the DataFrame is empty.


patudom:
You rock! Thank you for fixing this.

PR #27, opened 2022-10-08
Stage 3 ux 2

patudom:
I've made a bunch of UX edits to Stage 3. On my computer, this runs in the jupyter notebook, but it won't run in voila for some reason. I'm not sure if it's just my computer being funky, so I'm leaving this as a draft until @johnarban or @heywooddogwood can verify that it runs locally in voila for you.


patudom:
Note - I accidentally broke the Stage 1 -> Stage 2 Intro transition in this PR, so when reviewing, you can "Fill Data Points" on Stage 1 then jump straight to stage 3. I will send a fix to PR 31.

johnarban:
@patudom this PR looks good. 

I saw the Sage 1/2 transition break, but I went through and filled it in manually so I could see all the changes 

There are a couple things I've noted, just putting them here for now. 
> Stage 1: the reflection appears early on `rep_rem1` rather than on `ref_dat1` where it is introduced
> In Stage 3: we should toggle the line fitting and drawing tools off until the appropriate marker.

patudom:
Both @johnarban and @heywooddogwood were able to run this PR, so I will merge for now, and we can fix additional issues in later branches.

PR #28, opened 2022-10-11
S3 race dialog

heywooddogwood:
Converted the guideline alert for the running race guideline into a pop-out dialog. It still functions in the typical back/next structure of the guidelines. In the process, the multiple choice feedback items have been updated to get a semi-opaque black background to be easier to read on the guideline red. Also, the running race now has a split column structure that will allow for an image to be dropped in.


patudom:
Closing this since Harry has brought over needed features from this branch to PR #53.

Issue #36, opened 2022-10-14
Mixed up markers


patudom:
If you complete Stage 1 by going through the actual steps of the activity, when you start Stage 2, you will not be at the correct marker.

I think I figured out why this is happening, but I don’t know how to fix it:

- We use a lot of the same on_marker_update code in both Stages 1 & 2 & and presumably will in Stage 3 when we get that more built out.
- Markers that auto-advance in Stage 1 trigger this blurb of code in Stage 2:

`
 def _on_marker_update(self, old, new):`

        if new in self.stage_state.step_markers and advancing:  

            self.story_state.step_complete = True  

            self.story_state.step_index = self.stage_state.step_markers.index(
                new)`

This advances the Stage 2 markers through the "step_markers" sequence, so usually, you end up at "est_dist1" when starting Stage 2 instead of "ang_siz1".

`
    step_markers = CallbackProperty([  `

        'ang_siz1',  

        'ang_siz3',  

        'est_dis1'  

    ])

I guess this is happening because all the stage markers are listening for the same callbacks? Is there something we can do to make them independent of each other? Two of the markers are also repeated in Stages 1 & 2, so I tried to differentiate the Stage 2 marker names, but that didn't make a difference.

patudom:
Fixed by #40.

PR #37, opened 2022-10-14
Mixed up markers

patudom:
This is not meant to be merged in. I'm posting it in case it's helpful for reproducing what I'm reporting in Issue #36, which I think explains why we are having Stage 2 marker problems, but I don't know how to fix them.


patudom:
The relevant commit here is [439cb82] and possibly [5e04d58].

PR #38, opened 2022-10-14
All stages ux updates pat

patudom:
I believe Stages 1 & 2 and their intros should be more or less fully functional on a clean single pass through (i.e., as long as user doesn't log out and log back in or do anything crazy), except:

- See mixed up Stage 2 marker issue #36. @Carifio24, can you please help fix this before we deploy it this afternoon?
- Stage 2 doesn't automatically link to Stage 3, but I can fix that in the morning.
- There was a merge conflict from PR #30. I wasn't sure which should take precedence, so I accepted both and nothing has broken yet. Please fix if I've done that incorrectly.


patudom:
Closing this PR because it only has 3 commits. See instead PR #39. (I renamed this branch part way in, and it seems like it only sent the commits from prior to the name change).

PR #46, opened 2022-10-17
Allow selecting another galaxy if one is removed via flagging

Carifio24:
This PR fixes a bug (reported in the state testing Issue) where a student could not select replacement galaxies if they flagged one or more in stage two. Since the stage instances can't communicate between themselves except via a story, I fixed this using some glue messaging, which I thought was simpler than adding extra story state management and callbacks.


patudom:
I flagged a galaxy in Stage 2. When I returned to stage 1, the spectra were no longer loading:

Screen Shot 2022-10-18 at 12 31 25 PM

This may have been a fluke because when I tried again the spectra were there as expected. Will try a couple more times and see what happens.

patudom:
Regarding the above issue, it seems to happen if you go through all of Stage 1 and complete all the measurements. Every time I've done that (which is the likely scenario for a student), when I flag something in Stage 2 and come back, the spectra do not load.

Stage 2 is also behaving weirdly. Besides the large list of error messages sent to @Carifio24 on Slack (associated with the "next" button on the first Stage 2 guideline not advancing), the measuring tool did not appear as expected in the WWT viewer:
Screen Shot 2022-10-18 at 1 25 53 PM

patudom:
@johnarban and I decided for now to disallow flagging of galaxies. If they are able to display the galaxy to select in Stage 1, it should appear fully loaded in Stage 2, so hopefully they will never need to flag a galaxy. (And we think we've been thorough in purging bad spectra. And if they hit a dead tile on Stage 1, they just won't select it.)

patudom:
Closing this for now. We can resurrect it after the beta test.

PR #49, opened 2022-10-18
Stage 3 ux 4

patudom:
Lots of new Stage 3 guidelines, through the end of the reflection sequence for "my classmates' data."

- The guidelines are mostly in place, but much of the sequencing of available buttons/other logic doesn't work yet.
- Some reformatting of guideline components and editing of text still needs to happen.
- What viewers should appear with which guidelines is not correct yet. We need to think about what to make available. I think we may want to leave all previous viewers up once we get to them. Will need to experiment.


patudom:
Closing this. Superseded by PR #51

PR #51, opened 2022-10-19
Stage 3 ux 5

patudom:
This includes everything in PR #49 and more, so I am closing that one. @heywooddogwood, please review this one instead.

Lots of new Stage 3 guidelines, through the end of the reflection sequence for "my classmates' data."
- The guidelines are mostly in place, but much of the sequencing of available buttons/other logic doesn't work yet.
- Some reformatting of guideline components and editing of text still needs to happen.
- Suppress display of line-fit label until correct guideline is reached; for now display only age, not slope.
- Start experimenting with which viewers should appear as we pass through the guidelines. For now, most will stay visible "forever" once a certain guideline is reached, but the "OUR DATA" and the viewer with the slider are redundant, so replace the our data viewer with the slider viewer when that guideline is reached.


heywooddogwood:
Everything looked good in my run-through. merging.

PR #52, opened 2022-10-20
S3 slider styling

heywooddogwood:
Change color scheme to hard red to match the legend. Make the number in the thumb-bubble bold. Add faint red fill to the legend. Also change the labels to say "Low" and "High".


patudom:
The "Age (Gyr)" label I added looks correctly centered for an odd value of vmax. Will need to check behavior on an even value later.
Screen Shot 2022-10-20 at 10 25 39 AM

PR #59, opened 2022-10-21
Issues with no data

Carifio24:
This PR attempts to clean up some issues that will pop up if there isn't any class data present yet. This was done pretty quickly, so no guarantees that there won't be more issues.


Carifio24:
So there are still a couple of problems/non-ideal things here, but I have a beat on what they are. Will mark as ready once I fix them.

Carifio24:
This still definitely isn't perfect, but I think we might need these changes for things to run when there isn't any class data ready yet.

patudom:
This has now been solved by better handling of the situation where glue gets no data.

PR #62, opened 2022-10-22
Slides with images and text

LilyN11:


LilyN11:
All of Alyssa's text updates are on here. One thing I think we should address is captions for all the pictures. This is an accessibility question too so hopefully that can be added after initial testing. Additionally, Alyssa had big problem with the phrasing of Galaxy Physical Length. I took the liberty of creating a revised image for the angular sizes slide-show if we did want to change it. 
Screen Shot 2022-10-22 at 4 23 15 PM

patudom:
@heywooddogwood, I tried to put a caption under the WWT viewers, but when added it directly to the viewer columns (so it would be centered under the viewer), it ended up looking like this:  
Screen Shot 2022-11-11 at 12 40 45 PM

So I ended up putting the caption in a row under the whole sky viewer + instructions/buttons row, which makes it off-center from the WWT viewer.

PR #67, opened 2022-10-24
Add best-fit galaxy functionality

Carifio24:
This PR adds functionality to insert a best-fit galaxy into the student data. This best-fit galaxy has a distance value midway between the low and high distance's of the student's galaxy, with the velocity then determined by the slope. Once added, this galaxy appears in only the stage three table, and also shows up as different color in the "Our Data" viewer (This is done by making a subset that contains only that point. This point is actually also present in the student data layer, but if we set the opacity of the best-fit subset to 1, that's impossible to notice). Whether or not a student has added this galaxy is stored in the story state's `has_best_fit_galaxy` property. Note that this is NOT added as a measurement in our database.

The one thing that's missing is highlighting the table row with that color. I thought this would be easy to do, but the [`item-class` prop](https://vuetifyjs.com/en/api/v-data-table/#props) of the `v-data-table` was added in a later version of Vuetify that `ipyvuetify` uses.

### DIFF
--- a/src/hubbleds/components/age_calc_components/guideline_age_universe_estimate3.vue
+++ b/src/hubbleds/components/age_calc_components/guideline_age_universe_estimate3.vue
@@ -16,21 +16,8 @@
       v-intersect="(entries, _observer, intersecting) => { if (intersecting) { MathJax.typesetPromise(entries.map(entry => entry.target)) }}"
     >
       <p>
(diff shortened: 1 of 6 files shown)


patudom:
@Carifio24, can you please make it so that when the best fit galaxy is displayed, the best fit line is also displayed if it isn't already? Thank you!

Carifio24:
@patudom The layer viewer linefit tool should now automatically turn on when the best-fit galaxy is added.

PR #72, opened 2022-10-25
disable class layer toggle during line fit

johnarban:


patudom:
Closing because we already merged this in PR #70

PR #77, opened 2022-10-25
add pause before student uses the zoom tool

johnarban:
This PR forces the student to use the wavelength zoom tool before progressing.


johnarban:
Thanks, these changes look good @patudom

PR #79, opened 2022-10-26
add back the prodata viewer and new "pro_view" marker

johnarban:
added back the new prodata viewer. This is not properly sequenced. it shows up at the beginning of Stage 3, but it's guideline.

this can be fixed once i know the sequence


patudom:
Closing for now - will merge more carefully after beta

PR #87, opened 2022-10-27
Stage 3 updates v6

patudom:
Sorry for cutting things so close. On this branch, I think the flow from reading in stored data should mostly be ok. There is a chance that if students have previously entered freeresponses and they change them, the program may wig out. (I saw a bunch of error messages in the notebook that seemed like they may have been due to that, but it kept running locally).

There are a couple of logic issues - one I introduced and don't have time to fix, so I put a note to the students how to get around the issue.

The other is in John's pro_dat6 guideline. His logic looks correct to me, but the app is giving the opposite feedback, so I put in a note that there's a possibility the feedback is wrong. (Our class is teal; HST is lime green; Hubble is purple)
Screen Shot 2022-10-27 at 2 47 07 AM

### DIFF
--- a/src/hubbleds/components/age_calc_components/age_calc_components.py
+++ b/src/hubbleds/components/age_calc_components/age_calc_components.py
@@ -34,10 +34,12 @@ class AgeCalc(v.VuetifyTemplate):
         add_callback(self.story_state, 'responses', self._update_guesses)
 
     def _update_guesses(self, responses):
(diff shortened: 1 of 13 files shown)


Carifio24:
@patudom I've made a change so that updating the guess text from the state won't throw an error.

I'm not quite sure what's up with the pro_dat6 guideline - on my computer, the line colors are flipped. At least this isn't really a critical bug.

prodata

PR #90, opened 2022-10-31
Add index to keep track of progress with prodata components

Carifio24:
This PR looks to fix the issue where students could get stuck in the sequence of prodata components due to the reuse of the `prodata_response` variable in the stage state.

The approach used here is to have a `max_prodata_index` member of the stage state that keeps track of how far a student has advanced in the prodata sequence. Each component now has an internal `can_advance` flag as well as an index. The `can_advance` flag is set based on `max_prodata_index` and the component's index when it is created, and is now what controls the display of state in the template (rather than `state.prodata_response`).

I didn't look for any other places where the same issue could arise, but if there are, this can hopefully be a template for how to solve them.

### DIFF
--- a/src/hubbleds/components/prodata_components/guideline_professional_data0.vue
+++ b/src/hubbleds/components/prodata_components/guideline_professional_data0.vue
@@ -5,8 +5,7 @@
       state.marker = 'acc_unc1';
     "
     @next="
-      state.marker = 'pro_dat1';
-      state.prodata_response = false;
+      advance('pro_dat1');
     "
   >
     <div
--- a/src/hubbleds/components/prodata_components/guideline_professional_data1.vue
+++ b/src/hubbleds/components/prodata_components/guideline_professional_data1.vue
@@ -35,7 +35,7 @@
           :incorrect-answers="[0]"
           :correct-answers="[1]"
           :neutral-answers="[2]"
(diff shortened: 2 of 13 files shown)


johnarban:
So I was getting the following message running this PR -- it would persist after restarting the kernel. Once I changed branches (I switched to a branch where I had fixed the prodata sequencing, you can compare it [here](https://github.com/cosmicds/hubbleds/compare/main...johnarban:hubbleds:start-fixing-some-sequence-issues-with-prodata-section) yours is more extendable I think) it started working again, and then when I switched back it began working again. I am not sure what is causing it, but I see yours is watching the "can_advance" variable. This message only appeared if I had a user starting in the prodata section. 

image

johnarban:
@patudom - did you intend to skip guideline 3 in `guideline_professional_data2.vue`. I just notice it here, but it looks like you skipped it in [this](https://github.com/cosmicds/hubbleds/commit/e1503ccf65039bfb3d659f354326cbea3ead89f8#diff-d9cb1434bb35e005f078a2f1fed8a0894453131173ea8f33a0237825322bde81) commit?

patudom:
@johnarban yes, I decided at the last minute that it might be hard for the students to tell which had more random uncertainty. I missed removing the navigation to that guideline through the "back" button of the next guideline, so we'll have to remove that too. Thanks for catching that.

PR #91, opened 2022-11-03
Use visual attribute styling in new subset creation

Carifio24:
This PR updates the Hubble story to use new functionality from `glue-core` 1.6 (and thus relies on https://github.com/cosmicds/cosmicds/pull/174). As far as I can tell this only affects one location, but we may as well stay up to date. There probably won't be any apparent change from this update, but the goal is similar to https://github.com/cosmicds/cosmicds/pull/147 - prevent any color-changing flashes when creating a new subset with a specific color.

### DIFF
--- a/src/hubbleds/story.py
+++ b/src/hubbleds/story.py
@@ -283,11 +283,11 @@ class HubblesLaw(Story):
             if subset is not None:
                 subset.subset_state = subset_state
             else:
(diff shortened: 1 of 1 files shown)


johnarban:
this works! [cosmicds/cosmicds#174](https://github.com/cosmicds/cosmicds/pull/174) also merged

PR #93, opened 2022-11-06
Split stage 3 into three new sections

johnarban:
This PR splits stage 3 into three stages
 - (3) Explore Data (now ends on `sho_est2`)
 - (4) Class Data and Uncertainty (`ran_var1` - `acc_unc1`)
 - (5) Professional Data (`pro_datN` markers)

## Stage 3
The code of stage 3 remains largely unchanged, it mainly adds a new `_on_stage_three_complete` method and associated callback. This is triggered by setting the state variable `stage_three_complete` to true. Rather than changing the markers and `.vue` file, I trigger the `_on_stage_three_complete` method early on in [guideline_shortcomings_est2.vue](https://github.com/cosmicds/hubbleds/compare/break-stage-3-into-3-new-stages-and-add-measurement-of-same-galaxy...johnarban:hubbleds:split-stage-3-into-3-and-4-and-5?diff=unified#diff-56cae062498fd31991a6c1b78aa3552748fcadbda02244e31c453d4a3848a12a). The stage was also modified so that the `deferred_setup` method is run if we start in a later stage. So **this stage is still very heavy**.

## Stage 4
Stage 4 currently just pulls the viewers over from stage 3, in part because I kept running in to bugs and I wanted to get this out sooner rather than later, but also it will make it easier to shift things between stages. 

## Stage 5
The professional data series uses the method for advancing the slides I wrote in the [start-fixing-some-sequence-issues-with-prodata-section](https://github.com/johnarban/hubbleds/tree/start-fixing-some-sequence-issues-with-prodata-section) branch. I started using the fix by @Carifio24 in #90 , but I wanted to keep it completely self-contained and independent of the stage

## Known Issues
 - Stages transition forward, but you can go backwards using the "back" buttons
 - Stage 3 is very heavy
 - Viewers only in stage 4 should be created in stage 4 (currently in stage 3)
 - I did not fully reset all of the logic for which viewers were displayed together. On a run through every thing looks find, but like the issue below suggests, maybe not everything is fine
 - When you reach the first instance of the `layer_viewer` it shows two datasets, 
 ```
 (BqplotScatterLayerArtist for class_data, visible=False
 BqplotScatterLayerArtist for student_data, visible=True
 BqplotScatterLayerArtist for student_slider_subset_01, visible=True)
```
image

### Stage Test
Two additional files building from @Carifio24 test stage files, are meant to serve as templates for what needs to go into creating a new stage. Seeing those from @Carifio24 were super helpful, so I am trying to make ones that are a little more complete and _templatey_


patudom:
Looks good as a first step to breaking Stage 3 apart.

To do in a future PR:

- Remove "back" buttons from the first guidelines in Stage 4 & 5
- Move creation of viewers to the stages where they actually are used (instead of importing them all from Stage 3)
- Figure out where rogue subset that appears in all the Hubble viewers is coming from.

PR #96, opened 2022-11-08
Add guard against using beta test student ID

Carifio24:
This PR adds some logic that throws a `ValueError` if the student ID is between 2989 and 3017, which is the range of IDs used for the beta test students.

### DIFF
--- a/src/hubbleds/story.py
+++ b/src/hubbleds/story.py
@@ -429,6 +429,6 @@ class HubblesLaw(Story):
         super().setup_for_student(app_state)
         if self.student_user["id"] in range(2989, 3018):
             print("Warning! You are using the ID of a beta test student.")
-        app_state.update_db = False
+            app_state.update_db = False
         self.fetch_student_data()
         self.fetch_class_data()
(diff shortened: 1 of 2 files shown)


Carifio24:
@patudom Per our discussion today, I've changed this to set the app state's `update_db` value to false, rather than throw an error. The app also now prints out a warning.

src/hubbleds/story.py
```diff
@@ -427,5 +427,8 @@ def check_update(measurements):
 
     def setup_for_student(self, app_state):
         super().setup_for_student(app_state)
+        if self.student_user["id"] in range(2989, 3018):
+            print("Warning! You are using the ID of a beta test student.")
+        app_state.update_db = False
```
patudom, line 432: I think you mean for line 432 to be indented under the if statement?

patudom: Also, it looks like 2989 triggers the print warning, but 3018 does not, so the range is inclusive of 2989 but not 3018? I think that's what you intended, since 2989 is a test student and it looks like 3018 is not, but just wanted to make sure.

Carifio24: Yes! That was a bug in line 432. As for the range, 3017 is the last beta student, so 3018 should be the second argument to `range` (which is inclusive of the start, but not the end).

PR #100, opened 2022-11-08
tell users at beginning of stage 2 they can use menu to go to previous stages

johnarban:


patudom:
I edited the instructions slightly and added them also to the first guideline of Stage 1. I think between those 2, they should get the hang of the navigation menu and not need the instruction at the beginning of every remaining Stage. What do you think, @johnarban? It it looks good to you, feel free to merge.

johnarban:
I will merge this into main as well in PR #101

PR #102, opened 2022-11-14
Show previously selected data in selection tool

Carifio24:
This PR fixes a bug where a student's previously selected data (i.e. from a previous session using the app) wasn't showing up in the selection tool.

### DIFF
--- a/src/hubbleds/stages/stage_one.py
+++ b/src/hubbleds/stages/stage_one.py
@@ -219,7 +219,8 @@ class StageOne(HubbleStage):
         sdss_data = self.get_data(SDSS_DATA_LABEL)
         selected = self.get_data(STUDENT_MEASUREMENTS_LABEL).to_dataframe()
         selection_tool = SelectionTool(data=sdss_data, state=self.stage_state,
-                                       selected_data=selected)
+                                       selected_data=selected,
+                                       show_galaxies=self.stage_state.marker_reached("sel_gal1"))
         self.add_component(selection_tool, label='c-selection-tool')
         selection_tool.on_galaxy_selected = self._on_galaxy_selected
         selection_tool._on_reset_view = self._on_selection_viewer_reset
--- a/src/hubbleds/components/selection_tool/selection_tool.py
+++ b/src/hubbleds/components/selection_tool/selection_tool.py
@@ -43,14 +43,17 @@ class SelectionTool(v.VueTemplate):
         self.motions_left = 2
         self.gals_max = kwargs.get("galaxies_max", 5)
         
+        self.sdss_layer = None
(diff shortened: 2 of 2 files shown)


src/hubbleds/stages/stage_one.py
```diff
@@ -219,7 +219,8 @@ def __init__(self, *args, **kwargs):
         sdss_data = self.get_data(SDSS_DATA_LABEL)
         selected = self.get_data(STUDENT_MEASUREMENTS_LABEL).to_dataframe()
         selection_tool = SelectionTool(data=sdss_data, state=self.stage_state,
-                                       selected_data=selected)
+                                       selected_data=selected,
+                                       show_galaxies=self.stage_state.marker_reached("sel_gal1"))
```
johnarban, line 223:
```suggestion
                                       show_galaxies=self.stage_state.marker_reached("sel_gal1")  & self.stage_state.marker_before("sel_gal3"))
```

This addition makes sure the layer containing all the SDSS galaxies get's turned off. The current version shows the following (arrows show that the PR works and student galaxies show up):
image

With the suggested change, you see the following instead:
image

johnarban: @Carifio24 @patudom – This seems like what I would expect if I went back after going through some later stages. This data is from student 3327. I also thought that at some point this was the behavior and the selected points showed up in red (green+red=yellow). 

But I think w/ or w/o this change, it can be merged

Carifio24: I didn't see a place where the galaxy layer got turned off, but if I missed it then I agree it should be included in the initialization logic. (I'm also fine with modifying the behavior to turn it off at some point - the functionality to turn off the SDSS layer is already built in to the selection tool.)

johnarban: I don't _think_ I have seen that logic anywhere either, but I have at times seen just the red dots for the student selections (unless I dreamt it, or maybe it was a bug). But at present there isn't anything that I have seen which turns it off. 

So in that case, I think I am good to just merge what you had w/o the commit suggestion. Sound good 👍🏾 ?

Carifio24: Sounds good to me :thumbsup:

patudom: @johnarban, you are right that in the past, when a student logged out and logged back in, the green catalog dots were not displayed, and without them, the student's galaxies were displayed in red. That was a bug that didn't get fixed before the beta rollout, but I can't reproduce it now.

I will go ahead and merge this for now, with the behavior that the student's galaxies are yellow and the rest of the catalog appears as green. We can decide later if it would make sense for the green catalog dots to not be displayed when the student returns.

PR #103, opened 2022-11-17
Updates to axis limit resetting

Carifio24:
This PR contains some updates (and removes some unnecessary stuff) to make axis limit updating work. The updates here functioning correctly rely on https://github.com/cosmicds/cosmicds/pull/177.


patudom:
@Carifio24 The data update worked in my testing.👍 

We'll need to think about how to handle complicated situations like with this slider:

- The line fit needs to be updated.
- With the data update, the age value will be different, which may put it in a different sort order in the slider list.

Screen Shot 2022-11-30 at 1 45 17 PM

Carifio24:
Cool, glad to hear the data update is working  :thumbsup:

The slider already has logic to handle the data update - all we need to do is call its `refresh` method after that happens. This was in stage 3 pre-split, but I'll take a look at the new setup and make sure it gets called when appropriate.

I'll also take a look at the line fitting. The line fit tool should listen to changes in the data that it's fitting to, but maybe things are getting updated in a way that it doesn't expect.

johnarban:
Stage 4 & 5 were updated appropriately in #109

PR #111, opened 2022-12-05
Clean stage 3 and add layer togglers

patudom:
This PR removes the vestigial content from Stage 3 that had been migrated to Stages 4 & 5, adds the new layer toggler to the Stage 3 layer viewer and updates some of the Stage 4 behavior. (Basically when I cleaned out the Stage 3 content, it broke a bunch of things in Stage 4, so I had to patch them back up.)

Known issues:

- If you click "FILL DATA POINTS" in Stage 1 & then proceed to Stage 3, for some reason, the data points don't display until you click on the "classroom" icon. This doesn't happen if you measure the data in Stages 1 & 2, or if you prefill the data before loading the app calling `self.vue_fill_data() `in Stage 1.
Screen Shot 2022-12-05 at 12 21 12 PM
- There are various issues with the student open response boxes not keeping their content, but we can sort that out after we update the GlueState behavior.
- I haven't added the layer toggler to Stages 4 or 5 yet.


Carifio24 (approved):

Carifio24:
Looks good to me on a forward pass through. After looking at the guidelines again, I think the commit that I added may not have been the ideal way to fix things, but I'll fix that when I fix the stage 3 GlueState stuff.

PR #113, opened 2022-12-08
Updates to Stage 1, 2, 3

johnarban:
Updates Stages 2, 3 to use the new state management scheme where the state is passed via `:state` in the Vue component. 

Lock down various multiple choice questions in 1, 2, 3 to prevent advancing, and the Hubble Slide Show. I didn't lock the text feedback questions.


patudom:
Issues I've noticed so far:

- The last button of Stage 2 does not advance to Stage 3. (It appears to do nothing).
- If I collect Stage 1 & 2 data as normal (as a student would), the layer_viewer in Stage 3 does not have limits properly set. (It's at 0,0 and displays no data).

PR #116, opened 2022-12-16
Update Vue-only guidelines to use scaffold-alert

Carifio24:
This PR updates the Vue-only guidelines (the "generic state components") to all use the `scaffold-alert` component. This doesn't touch any of the slideshows or guidelines with a Python component - I'll handle those in another PR soon.

I also rearranged these files to account for what stage they're in post-Stage 3 split. If I've put one in the wrong folder, let me know and I'll change it.

Note that this shouldn't depend on https://github.com/cosmicds/cosmicds/pull/186 - I made the changes there with an eye towards flexibility, but nothing here should actually need anything new from there.


Carifio24:
I've changed all of the `header-text` prop assignments to be `title-text` instead since the changes in https://github.com/cosmicds/cosmicds/pull/188 will show the header if any text is present. This isn't actually a prop that `scaffold-alert` looked for, but I wanted to not show the headers while retaining our old titles in case we change our minds on using them. (It's one shell command to change these back to `header-text`).

PR #117, opened 2023-01-03
Updates to slideshows, components, and guidelines

Carifio24:
This PR makes updates to our slideshows and components to minimize references to state inside the components, and pass in state as a prop when necessary. This PR also moves all of our guidelines to use our general `scaffold-alert` component and strips out any remaining Python-side guideline pieces.

For the professional data component sequence, I created a `guideline-professional-data` component which all of the sequence guidelines use as a base - hopefully this serves as an example for how to do something similar to other guideline sequences. In order to correctly know (and manage) whether or not the user can advance, these guidelines depend on https://github.com/cosmicds/cosmicds/pull/189. This PR also makes changes to reflect the change to the multiple choice component made there.


patudom:
Thanks, @Carifio24 - these PRs include a lot of important changes!

I've noticed 2 issues so far. I don't know if the first needs to be fixed here or in the cosmicds repo, but I will note it here:

- The mc score is showing up prematurely by the piggy bank icon in the upper right of the app. (If you keep choosing wrong answers, the point total decrements as it should, but it shouldn’t be displayed until the student gets the right answer.)
Screen Shot 2023-01-06 at 6 07 27 PM



- Measuring the galaxy angular size in Stage 2 (marker ang_siz4) does not allow the guideline to advance (so I can't check anything else beyond this point).
Screen Shot 2023-01-06 at 6 23 44 PM

Carifio24:
@patudom The issue with not being able to advance from the measuring distance should be fixed now (was due to a typo in the relevant callback). I think the scoring issue may be a cosmicds (rather than hubbleds) issue, so I'll take care of that in a separate PR.

PR #118, opened 2023-01-04
Snackbars to nudge students forward from WWT widgets

Carifio24:
This PR adds functionality for prompting students to move on from playing with the WWT widgets after some time has passed. I've marked this as a draft so that @heywooddogwood can finish styling/adjusting content of the nudging content itself (which is currently just a simple snackbar).

We keep track of which timers have started and finished in the `timer_started` and `timer_done` dictionaries inside the intro slideshow. The relevant timer for each of the slides with a WWT widget is started when the student first pans or zooms in the WWT widget, or first presses one of the location buttons on that slide. The duration between the first WWT interaction and the snackbar appearing is set by the `timer_duration` value (specified in milliseconds). I've currently set this to 5 seconds for ease of testing, but obviously we'll want it to be longer than that.


patudom:
Thanks, @Carifio24! Functionality on the timers seems to be working well.

@heywooddogwood, I would request that the popup appear closer to the WWT window (or maybe even right on top of it). I didn't notice the snackbar that appeared at the bottom of the slide at first.

PR #119, opened 2023-01-04
Various updates 1

heywooddogwood:
This pull request references https://github.com/cosmicds/cosmicds/pull/190 in cosmicds repository, and updates the styling in the stage 1 scaffolds. widget button references now use the green-circle icons as they appear in the widgets. Same for references to blue buttons in toolbars (although in both cases, we only currently have one or two examples of each, the first time or two that they appear in the scaffolds). Text is set to white in the footers, and left-side border bars highlight text instructions when next/back buttons don't immediately appear. Much of the styling is standardized and controlled by the scaffold_alert.vue component contained in the cosmicds repository.


patudom:
Thanks, @heywooddogwood! This is looking much improved.

Aside from the issue we discussed where the text on the guidelines needs to be adjusted to pure white in dark mode and black in light mode, I noticed a couple other things:

- I pushed a fix where the bottom button row on the one of the guidelines was repeated by accident.
- I don't know if this is intentional, but I think maybe some of the styling that was intended for the text input boxes is somehow being applied to the multiple choice text and the age slider, which are now appearing brown to me. Sorry if it is intentional, but I'm not a fan of this color scheme and hope we can change it.
Screen Shot 2023-01-06 at 7 59 35 PM
Screen Shot 2023-01-06 at 8 08 18 PM

PR #123, opened 2023-02-23
Fix issues in stages 2, 3, and 4

Carifio24:
This PR fixes some post-state-migration issues, primarily in stage 3:
* The limits on the "Our Data" viewer should now be set correctly
* Fix issue where the "Hubble's Discovery" button wasn't appearing (which prevented advancing)
* Make some updates to the stage 3 layer toggler

If there are any issues with the sequencing, etc. let me know - I wasn't 100% sure how this section is supposed to go, but this PR fixes the main technical issues. Note that this relies on https://github.com/cosmicds/cosmicds/pull/193.

Additionally, this PR swaps out the deprecated `DataFrame.append` call in the selection tool with `pandas.concat`.

### DIFF
--- a/src/hubbleds/components/__init__.py
+++ b/src/hubbleds/components/__init__.py
@@ -1,3 +1,4 @@
+from .angsize_dosdonts_slideshow import DosDontsSlideShow
 from .distance_sidebar import DistanceSidebar
 from .distance_tool import DistanceTool
 from .exploration_tool import ExplorationTool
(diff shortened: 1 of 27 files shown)


Carifio24:
I was originally going to try and tackle stages 3 and 4 separately, but there ended up being some crossover between the changes, so I think it's easiest to handle both stages in one PR.

Carifio24:
Okay, I think this is ready now.  I'm sure there are still some minor sequencing issues, but this should get stages 2-4 back to the point of usability (I was able to go through them all moving forward).

PR #133, opened 2023-03-23
Spectrum measurement tutorial sequence

johnarban:
This PR adds in a new sequence for remeasuring the galaxy. This adds a new table `example_galaxy_table` and vue component `SpectrumMeasurementTutorialSequence`. The vue component is `c-spectrum-measurement-tutorial`

In stage one, several changes callback functions have been made to make them more general: 
1. `_on_galaxy_row_click` &rightarrow; `on_table_row_click` and accepts the table associated with the row as a parameter. this way it can be used for multiple tables. It is wrapped in a `lambda` function when used in a callback
2. `galaxy_table_selected_change` &rightarrow; `table_selected_change`: The `change` dictionary, provided by the hub observer, includes an `owner` key which includes the table. This is now used instead of directly referencing one of the tables.
3. A new callback has been added for `on_spectrum_click_example_galaxy` which adds the ability to fill in the galaxy velocity if we have reached that point in the tutorial. The problem here was I couldn't find a way to plug directly into whatever was automatically updating the velocities for the selected galaxies. 

Some problems:
- Newly added state variables are being reset to their original value. These include `show_meas_tutorial`, `show_galaxy_table` and `show_example_galaxy_table`. Right now, visibility is being controlled with markers
- Histogram - The outlier points are no longer clear because the bar height (or if we used the dotplot, the marker size) is too small. It currently only shows a single bar clearly when zoomed all the way out.  image </br> Zoomed in view image
- **This only pulls data from the database** so far as I know, it is not hooked up to send it back. This is fine for now.


Carifio24:
We discussed the ability to filter in tables yesterday, so I've got an implementation of that in https://github.com/cosmicds/cosmicds/pull/197 if it's useful for this sequence. I tried this branch out on my machine yesterday and I'm also seeing the issue with the new callback properties getting reset. Still trying to get to the bottom of that.

patudom:
Merging so we can proceed with the file/stage name changes. I can provide more detailed feedback after that process is complete. Thanks!

PR #146, opened 2023-04-08
Add Layer Toggle to Stage 6 and various guideline edits

patudom:
- Add layer_toggle to Stage 6
- Various guideline text and styling edits


heywooddogwood:
Checked layer toggle on my system, seems to be working properly.

PR #150, opened 2023-04-11
spectrum measurement tutorial seq

johnarban:
This PR contains the completed Spectrum Measurement Tutorial sequence in Stage 1.  It has been merged with the most recent version on main. I have added in some sequencing steps. 

Sequencing can be a bit complicated by the fact that guidelines for this are in a self contained vue only component. This means user interactions from inside `guidelines-spectrum-measurement-tutorial` can't access the state. To get around this, we `emit` back to the parent, and then set variables or run things from there (unless the interaction is internal to the component). See [here](https://github.com/johnarban/hubbleds/blob/e5f4cc6ab5d83725949d0800916ef673fe5a35ad/src/hubbleds/components/generic_state_components/stage_1/guidelines_spectrum_measurement_tutorial.vue#L49-L64)  for an example of both of these.  

The tower selection functionality is housed within a tool in `bin_select.py` called `SingleBinSelect`, it just creates a subset state, and then the selection is handled in the component. 

This PR technically depends on a branch of cosmicds here https://github.com/johnarban/cosmicds/tree/DotPlot-fixes that forces the opacity to 1 (it wouldn't change for me by just setting `alpha=1`) and more importantly a change so that it determines the y-axis bounds strictly using visible layers. The PR works fine without it. 

There are some minor changes to just show the dotplots in Stage 3, but I am working on that section next.

There are a couple things I did not figure out, that perhaps @Carifio24 could help with. 
1) is making sure it is doing things efficiently. I ironed out some things, but the veiwers can sometimes be jumpy with axes changes and it can get slow if it sits a while. Also, I was unable to link the y-axes of the dotplots in such a way that it is always the max of the 2.
2) We wanted to be able to remove the subsets on reset, however, when I tried, it ended up getting errors if I tried to use the `subset.remove()` command. I am unsure the right way to safely remove it. Maybe we can just hide it?
3) I also don't think this writes back to the database. There are a couple skeleton functions, but nothing actually done. Mainly, it's just daunting cuz I don't wanna screw up the database.  

And @patudom - I have reworded and reordered some of the guidelines, and tried to put user interactions in reasonable places. Between vue functions of the some of the state variables it should be possible to control just about anything. 

I am happy to add more notes here!


src/hubbleds/viewers/hubble_dotplot.py
```diff
@@ -0,0 +1,66 @@
...
+__all__ = [ 'HubbleDotPlotView', 'HubbleDotPlotViewState', 'HubbleDotPlotViewerState']
+
+
+class HubbleDotPlotViewerState(LineHoverStateMixin,DotPlotViewerState):
+    
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        
+
+class HubbleDotPlotViewState(LineHoverViewerMixin,BqplotDotPlotView):
```
Carifio24, line 16: This shouldn't have `State` in the class name, since it's not a state class, it's a viewer class (that ultimately gets wrapped by `cds_viewer`)

src/hubbleds/tools/bin_select.py
```diff
@@ -0,0 +1,144 @@
...
+
+        super().__init__(viewer, **kwargs)
+
+        self.interact = BrushIntervalSelector(scale=self.viewer.scale_x,
+                                              color=INTERACT_COLOR)
+
+        self.interact.observe(self.update_selection, "brushing")
+
+    def update_selection(self, *args):
+        with self.viewer._output_widget or nullcontext():
```
Carifio24, line 46: Just curious, why do this and `activate` have context managers? The purpose isn't immediately clear to me.

src/hubbleds/tools/bin_select.py
```diff
@@ -0,0 +1,144 @@
...
+            bins = self.viewer.state.bins
+            bin_centers = (bins[:-1] + bins[1:]) / 2
+            if self.interact.selected is not None:
+                x = self.interact.selected
+                if x is not None and len(x):
```
Carifio24, line 51: No need to check if `x is None` since it was just done above.

Also, I imagine that `len(x) == 2` so it doesn't really matter, but I think it's better to compute `min(x)` and `max(x)` once and avoid repeating the computation several times.

Carifio24 (changes requested):

Carifio24:
Very cool sequence! I'll take a look at some of the particular issues you mentioned, but I have just a few quick comments that I left above.

patudom:
- When activating the zoom in tool, the gray measuring tool should disappear.
- After a zoom in on the dot plot, all the tools should get reset
- Update size of "me" dot to match the size of the dots in the viewer
- If "me" dot is not in range being displayed, fix bug where the viewer tries to put the "me" dot into a present bin anyway.

patudom:
Closing this per Slack conversation with @johnarban. All these commits are already in #156, so will work from there.

PR #153, opened 2023-04-13
update distance and velocity to use constant variable names 

johnarban:
When adding measurements, `stage.py` would raise an error inside `update_data_value` because `distance` and `velocity` were not components. I updated this to use the constant values `VELOCITY_COMPONENT` and `DISTANCE_COMPONENT`


Carifio24:
Looks like we both had the [same idea](https://github.com/cosmicds/hubbleds/pull/148) :+1:

patudom:
Closing this because #148 includes this same update.

PR #155, opened 2023-04-14
Downselect number of seed galaxies

patudom:
This includes @johnarban's code for randomly choosing a subset of example galaxy seed measurements from the database (because we currently have too many and the velocity dot plots look like lines rather than dots when all data points are included).

We should adjust the value of "N" to something that works well for both the velocity and the distance dot plots.

I also added skew & rotation for the dot plot points, but it's possible we can remove these and make the points circles, since reducing the number of data points makes the dots larger and more distinctive.

ETA: This looks like a lot of commits, but this branch includes all of John's commits from #150, so once they are both merged, this will only be 3 commits (I think).


Carifio24:
Just a quick note that I've updated the random choice selection to use the current numpy pattern for random generation, rather than the legacy form which uses a singleton state behind the scenes.

PR #156, opened 2023-04-20
Stage 3 updates

johnarban:
This brings together everything we need for stage 3.

It should be merged after stage 2 (PR #150) , and is current with main (as of b87d094ddc30740da76a872d67ae002551586e43 / apr 20)


patudom:
There are a few issues that I noted in the punchlist, and as @johnarban and I have discussed, we will be streamlining a lot of the Stage 1 tutorial content. I'm merging this in for now because I realized I merged #155 to main without the update in [1819fda], so what's currently on main will probably not run without this PR. 

There are also some errors that show up in the notebook, but they don't seem to affect the user experience, so we will clean all that up hopefully tomorrow.
(https://github.com/cosmicds/hubbleds/pull/156/commits/1819fdad9e3f534a3fa172a6e893afef97faf042)

PR #157, opened 2023-04-21
"Waiting screen" for stage 4

Carifio24:
This PR adds a "waiting screen" for students who reach stage four before "enough" of their classmates finish collecting their data. Currently, the waiting screen allows them to use a WWT window until things are ready for them to proceed. The condition for this is connected to the class data - when enough students have finished taking measurements (that is, they have 5 data points), the waiting screen is taken away. Currently, I've defined "enough" students to be 6. We can easily change that - it's set on line 461 in `story.py`.

To test this, it's easiest to turn off the automatic class data updating to have some good control over what the class data actually is. Here's what I did:

* Comment out lines 144-145 and 483 in `story.py` (where the repeated timer is created, and where we initially fetch the class data).
* Manually set the class data to include data from 5 students via:
```
import requests

from cosmicds.utils import API_URL
from hubbleds.utils import HUBBLE_ROUTE_PATH

story = app.story_state
n_students = 5
n_measurements = 5 * n_students
url = f"{API_URL}/{HUBBLE_ROUTE_PATH}/stage-3-data/3005/178"
res = requests.get(url).json()
measurements = res["measurements"]
measurements = measurements[:n_measurements]
data = story.data_from_measurements(measurements)
data.label = "class_data"
app.data_collection["class_data"].update_values_from_data(data)
```

You should see the loading screen. Re-running the above block with `n_students = 6` should update the stage to show the regular content.

### DIFF
--- a/src/hubbleds/components/generic_state_components/stage_4_waiting_screen.vue
+++ b/src/hubbleds/components/generic_state_components/stage_4_waiting_screen.vue
@@ -12,7 +12,7 @@
       <v-toolbar-title
         class="text-h6 text-uppercase font-weight-regular"
       >
-        Waiting Screen
+        Take a quick break
       </v-toolbar-title>
       <v-spacer></v-spacer>
       <speech-synthesizer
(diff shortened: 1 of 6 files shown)


johnarban:
So this works. I think adding in the brightness/contrast slider might be good. One could a ton of little things, but I think the brightness/contrast adjuster would be enough for now. 

I found one thing, I don't know if the situation is possible in _real life_, but it is possible to get into a state where the "next" button on the last guideline of stage 3 (`fil_rem1`) doesn't advance. 

You get into this state by proceeding through with `n_students = 5`, and then getting to stage 4, and increasing to `n_students = 6`. Everything works ok still. But then if you set `n_students = 5` again, and go back, in stage 3 the `next` button stops working (you may have to go back and forth twice). The menu item still advances you.  And then, if you manually set `enough_students_ready = False`, and click the menu for stage 4, it goes to the right place, but if you go back to stage 3, clicking next advances you backwards to `est_dist2` (I don't know why). 

I think this is an impossible condition to run in to in the real world since students can't remove measurements. So these are ok.

PR #160, opened 2023-04-25
Fix state issue with angular size measurement

### LINKED ISSUE #159: State bug for example galaxy angular size measurement
In Stage 3, at marker ang_siz3, the user needs to measure the angular size of the galaxy to advance, which it does correctly. If you move backwards to the ang_siz3 guideline, the state doesn't register that the angular size has already been measured, and it is not possible to go forwards.

Carifio24:
This PR fixes #159. The updates here add a missing stage state variable, and update stage state variables in response to measuring tool updates.


patudom (commented):
Thanks! This seems to be working fine, so I'm assuming we can remove the console.logs?

Carifio24:
Yeah, the `console.log`s can be deleted.

PR #162, opened 2023-04-25
Add measurement tutorial state to tracked state variables in stage 1 (and fix sequencing issue)

johnarban:
This makes sure the current state of the measurement tutorial is tracked by state variables saved in the database. When it initializes, it goes through each step to make sure appropriate viewers are loaded.

fixes an issue with where `selected-callback` is used when the new method is to use `@selected` in `mc-radiogroup`


patudom:
Per Slack conversation with @johnarban, we are going to merge this in, so we have all the backend pieces we need for an upcoming overhaul of the Stage 1 tutorial sequence.

src/hubbleds/components/spectrum_measurement_tutorial_sequence/spectrum_measurement_tutorial.py
```diff
@@ -169,7 +180,14 @@ def __init__(self, viewer_layouts, *args, **kwargs):
...
         
-            
+        # run through steps to open run necessary setup calls in the vue file
+        for i in range(self.maxStepCompleted):
+            self.step = i
+    
+    def _on_tutorial_state_change(self, change):
+        self.print_log(f'name: {change["name"]}, old: {change["old"]}, new: {change["new"]}')
+        self.saving_state.update({change['name']:change['new']})
+        self.tutorial_state = self.saving_state.copy()
```
Carifio24, line 190: Are `tutorial_state` and `saving_state` ever used differently? I guess I'm not sure why there are two dictionaries with the same content.

johnarban: I did this because otherwise it seemed like the corresponding `spectrum_tut_state` in `stage_1.py`wasn't realizing that it had been changed somewhere else. It worked with this approach, so this just seemed safer at the risk of perhaps _poor_ style or extra steps.

PR #163, opened 2023-04-26
matching to bin center now works outside bin range

johnarban:
measurements, when added to histogram, can lie outside the bin range.


patudom:
Thanks, @johnarban. This is working well for measurements that are greater than the largest seed value, but I think it needs to be extended to include measurement values that are less than the smallest seed value.

patudom:
Also, I think the same functionality needs to be applied to the angular size and distance measurement dot plots too, in case the measured values fall outside the range of seed data values.

johnarban:
Ahh, so it does work for values less than the smallest seed values, but the limit reset doesn't respect the presence of the mark. I will take a look at this

johnarban:
So the problem seems to be the `HistogramViewerState` only looks at data in the layers and not at all of the marks. I don't think the fix for this belongs in this PR. Since this PR does work as intended, I think it can be merged. 

A couple of solutions to the problem are to readjust the x-limits (via `extend_tool`) after a mark is created or to create a new `Data` object, that way the viewer "sees" the mark.

PR #164, opened 2023-04-26
cleanly close spec measurement tutorial

johnarban:
Make sure the spectrum measurement viewer opens and closes with only the appropriate marks shown


patudom:
Per Slack conversation with @johnarban, we are going to merge this in, so we have all the backend pieces we need for an upcoming overhaul of the Stage 1 tutorial sequence.

PR #166, opened 2023-04-26
Fix total score in final guideline

### LINKED ISSUE #145: Final guideline in Stage 6 does not display correct score total
guideline_story_finish is supposed to display the student's piggybank total, but it looks like the story_state is not being passed correctly.

(If you click the "state" button on Stage 6, it lists total_score and other values as 0.)

Screen Shot 2023-04-07 at 5 26 06 PM

Carifio24:
This PR fixes the display of the total score in the final guideline. The issue is that the story state doesn't get updated in the frontend - I'm thinking this may be related to the issues we were seeing in https://github.com/cosmicds/cosmicds/issues/176. This hadn't surfaced before because this is actually the only place that we've tried to use the story state in the frontend, at least in the stages.

This PR provides a fix by adding a `total_score` member to the stage 6 state which is synced to the story state value.

Fixes #145.

### DIFF
--- a/src/hubbleds/components/generic_state_components/stage_6/guideline_story_finish.vue
+++ b/src/hubbleds/components/generic_state_components/stage_6/guideline_story_finish.vue
@@ -13,7 +13,7 @@
       class="mb-4"
     >
     <p>
-      You have completed the Hubble Data Story with {{ story_state.total_score}} <v-icon>mdi-piggy-bank</v-icon> points. Nice work!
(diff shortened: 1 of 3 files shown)


patudom:
I'm not sure why this change would cause this behavior, but I can't advance past the first mc question on this branch (but I can on other branches).

Screen Shot 2023-04-26 at 5 12 21 PM

Carifio24:
Huh, that's really weird - I'm seeing the same thing though, so it's definitely something in this commit. I'll investigate.

patudom:
Closing for now until we have a chance to investigate further

PR #168, opened 2023-04-26
Markers align to binning

johnarban:
Make a change in stage 3, so that when clicking in one plot and placing a marker, the corresponding marker in the other plot goes to the nearest bin. 

This builds off of PR #163 

https://user-images.githubusercontent.com/7862929/234689649-49974cb9-bab1-4d51-98b6-e969da0fa019.mov


patudom:
Per Slack conversation with @johnarban, we have decided to stick with the algorithm we already have. Thanks for trying this anyway!

PR #181, opened 2023-04-27
Create front end for simplified dot sequence in Stage 1

patudom:
- This creates the guidelines for a simplified tutorial sequence in Stage 1
- Back end actions or front end slideshows that still need to be created are noted direction on the guidelines in light green. We will need to remove those once the back end content is complete.
- I realized that the guideline text I wrote describing consensus only applies if students don't measure a far to the right outlier, so I will need to modify the text to be more general. I'll do that on a future pass.

@johnarban, let me know if this looks ok to you, and if so, you can start hooking up the back end functionality and I can create the shorter tutorial slideshow.


patudom:
Closing because everything here is in #186 already.

PR #184, opened 2023-04-28
Create backend for new stage 1

johnarban:


patudom:
Closing because everything here is in #186 already.

Issue #188, opened 2023-04-28
Stage 6 pro_dat6 needs to give correct feedback


patudom:
<img width="1310" alt="Screen Shot 2023-04-28 at 4 20 57 PM" src="https://user-images.githubusercontent.com/12750048/235246005-0317cd3d-72ec-4e14-9369-eb20c9fb9e5e.png">

johnarban:
This seems to happen because `stage_state.our_age` is never set in the stage.

PR #192, opened 2023-04-29
Fix stage 6 feedback (issue #188)

johnarban:
Fixes issue #188

### DIFF
--- a/src/hubbleds/components/generic_state_components/stage_6/guideline_professional_data6.vue
+++ b/src/hubbleds/components/generic_state_components/stage_6/guideline_professional_data6.vue
@@ -24,9 +24,7 @@
             'Our class got a higher age estimate than HST',
             'Our class got a lower age estimate than HST',
           ]"
-          :feedbacks="(parseFloat(state.hst_age) < parseFloat(Math.round(state.our_age).toFixed(0))) ?
-  ['Correct! The slope of best fit for our data is shallower than it is for the HST data', 'Try again! Compare the slope of best fit for our data to the slope for the HST data'] :
-  ['Try again! Compare the slope of best fit for our data to the slope for the HST data','Correct! The slope of best fit for our data is steeper than it is for the HST data', ]"
+          :feedbacks="feedbacks(parseFloat(state.hst_age) < parseFloat(Math.round(state.our_age).toFixed(0)))"
           :correct-answers="(parseFloat(state.hst_age) < parseFloat(Math.round(state.our_age).toFixed(0))) ? [0] : [1]"
           :neutral-answers="(parseFloat(state.hst_age) < parseFloat(Math.round(state.our_age).toFixed(0))) ? [1] : [0]"
           @select="(status) => { if (status.correct) { allowAdvancing(); } }"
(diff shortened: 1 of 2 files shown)


patudom (changes requested):
There are a few issues with this.

1. The question is how the HST key project result compares with our **class**'s data (not "my data"), so the relevant calculation is for the CLASS_DATA_LABEL, not student.

2. On the front end, we round the ages to integers to indicate that our method of estimating the age is not refined enough to have better than integer uncertainty, but as far as this backend calculation goes, we need more decimal places, given how close the slopes of these two lines might end up being. (so state.hst_age needs to be adjusted).
Screen Shot 2023-04-30 at 9 37 49 AM

3. Actually, rather than converting slopes to ages and risking round off errors, I would suggest that we fit a line to the HST key project data (as it is not changing) and just store the slope in our app as a constant with multiple decimal places. Then when we calculate the slope for "my class"'s data, we can just compare that against the HST key project slope. 
4. One potential issue is what to do if someone is so far ahead of others that the class's slope keeps changing as more data come in. @Carifio24, can you advise on a plan for managing that, please?
5. I don't see where in the code hst_age_less_then_our_age is being set.
6. This is my bad, but for this guideline, we should turn off the "my data" and "Hubble data" layers, so the graph is not overwhelmed with 4 fit lines in multiple colors.

johnarban:
I can make these changes this afternoon.

`hst-age-less-than-our-age` is given by the result of the long Boolean expression in the `:feedback` part of the guideline

And yes, I saw the issue with the HST age, especially since we only give them higher or lower, and if it is an integer it could be the same

johnarban:
Alright, I have done I think each of the items mentioned here

- [ ] hard code HST age = 12.79687910 Gyr
- [ ] switch to the class age (rounded to 2 decimals)
- [ ] show the class and HST key data on pro_dat6
- [ ] these should update as the data updates. I didn't do anything to limit the rate at which it updates. Hopefully, with pause screen, no one's data should be updating to much

patudom:
Thanks so much, @johnarban! I wanted to confirm the behavior with a different set of class data points, so I tried @Carifio24's trick from #157 to use a different batch of class data.

It looks like the class slope in the graph is still for the default batch of class data. Is this because the code in #157 triggers things differently than for an actual class in progress, or do we need to worry about whether we will be getting the correct batch of class data here? (i.e. related to bullet 4 in my original comment at the top of this thread).
Screen Shot 2023-05-01 at 8 34 03 AM

johnarban:
hmmm.....it may be that the way #157 updates the data doesn't trigger a message to be sent. you could, but it is just reading out of `class_data` and updates the slope when it changes.

Carifio24:
The snippet in #157 should trigger a `NumericalDataChangeMessage`, which is what stage 6's `_on_class_data_update` is listening for. I'll pull this down and take a look.

Carifio24:
So after looking at this, I can confirm that:
* The stage state's `class_age` updates when the class data is changed
* The drawn line updates when the class data is changed
I used the snippet from #157 to change the data

Maybe this set of data points just has a slope that's similar to that from the "default" class?

PR #194, opened 2023-04-29
Update y limits on spectrum viewer zoom

### LINKED ISSUE #183: Reset y-axis limit on spectrum viewer on zoom-in
The "trio" of lines at the h-alpha marker do not really look like lines when you zoom way in because of the y-axis stretch. Can we call a reset_limits() to fit what's displayed in the zoom?

So instead of looking like this,
Screen Shot 2023-04-28 at 9 52 50 AM

On zoom in, it could look like this:
Screen Shot 2023-04-28 at 10 13 35 AM

Carifio24:
This PR aims to resolve #183 by resetting the y-limits of the spectrum viewer when the speczoom tool is used.

### DIFF
--- a/src/hubbleds/tools/wavelength_zoom.py
+++ b/src/hubbleds/tools/wavelength_zoom.py
@@ -22,6 +22,8 @@ class WavelengthZoom(BqplotXZoom):
             return
         super().update_selection(*args)
 
+        state._reset_y_limits()
+
         if self.on_zoom is not None:
             xbounds_new = [state.x_min, state.x_max]
             self.on_zoom(xbounds_old, xbounds_new)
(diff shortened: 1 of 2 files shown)


patudom:
Thanks, @Carifio24! This is a super picky thing, which we can leave as is if it's a pain to change, but something about the reset is making the original zoom level have a much higher y-axis max than it used to. Is it possible to use the original y_limits when we first display the spectrum and when the user clicks "bqplot:home"? 

Screen Shot 2023-04-30 at 9 55 14 AM

Carifio24:
Hmm...I'm not able to reproduce any limits this far off. What galaxy are you looking at in this screenshot?

patudom:
I didn't think to write down the galaxy number. This one might be particularly weird because there is another line that is significantly brighter than the H-alpha. I tried randomly checking a bunch more and never came across another one this strange, so I will go ahead and merge it. THanks!

PR #195, opened 2023-04-29
Fixes stage 3 dotplot viewer x range

johnarban:
This PR fixes cosmicds/cosmicds#220

However, there is an additional issue where if the points are too small or too large `reset_limits` doesn't include them. 

The last commit (4a9851a8ca5062074b3d43e20c1217b41e98b1fb) fixes this - but if the angular size is too small (say 1 arcsec) the axes are thrown out of whack). For reasonable values it works well.


patudom:
@johnarban, it looks like in both the 1st and 2nd measurement dot plots, "my" 2nd measurement is being plotted in both (instead of plotting my 1st measurement in the first dot plot)

Screen Shot 2023-05-02 at 11 44 16 AM

patudom:
Also, I thought we had rounded all the labels to integers but that isn't the case here. Don't know if I imagined that we changed them before or if somehow they got put back. (I am looking at a version where the current main is locally merged into this branch). Not the most urgent fix, but would be good to update if it would be easy.
Screen Shot 2023-05-02 at 11 53 42 AM

johnarban:
both of these are fixed. 

measurements no update when changed, and the appropriate measurements is shown each plot, and the home button includes the far out points

PR #198, opened 2023-04-30
Make wavelength zoom aware of x-axis bounds

Carifio24:
I realized that #194 didn't actually solve the main issue of #183, since the `_reset_y_limits` method called there isn't zoom-aware. This PR updates the wavelength zoom tool to reset the y limits based on the x-axis bounds. @patudom let me know what you think about the behavior of this.


patudom:
Yes, this looks great! Thank you.

PR #202, opened 2023-05-01
Finish new stage 1 tutorial 3

patudom:
Some more updates to front end:
- add guideline to think about velocities
- split dot_seq13 into 2 parts - 1 to click row and 1 to make measurement
- small updates to where components appear in stage_1.vue
- Remove unused spectrum info button

Issues in my code:
- I added new state variables to trigger whether you can advance on dot_seq13 and dot_seq13a, but they are not hooked up yet. I tried to initialize them to true until someone else could hook up the back end, but I think we are running into the flaky state variable situation where one of them appears true and one of them appears false when you click state, even though they should both be set to true. (And when the dot_seq13 variable was seen as true in the "state" console.log, I still wasn't able to advance on that guideline).
- If we can't get those variables working properly, we can just leave them as is and let the user be confused if they don't follow the instructions.


patudom:
Should be fixed:

After 1st example galaxy spectrum measurement is made (before int_dot1), the dot plot and spectrum values need to be reset to full range. (They currently emain at whatever zoom in the student made when measuring the galaxy)
Screen Shot 2023-05-01 at 9 53 01 AM

patudom:
Should be fixed:

At dot_seq13, the 2nd measurement wavelength value in the table should be blank (not 0.)

patudom:
Should be fixed:

Any time the dot plot viewer makes the spectrum viewer zoom, the spectrum viewer should call the reset_limits from #198. (If you zoom in spectrum viewer, it zooms correctly, but if you zoom in dot plot and the spectrum viewer zooms with it, it uses the old zoom).

This is what it should look like:
Screen Shot 2023-05-01 at 8 49 15 AM

This is what it currently looks like:
Screen Shot 2023-05-01 at 8 48 51 AM

patudom:
Fix if possible:

johnarban:
> * dot_seq13

I have hooked these up, and they seem to be working for me. If they break for you, then you can set them back to false w/o any consequences.

johnarban:
All of these notes have been completed

PR #204, opened 2023-05-01
Replace toggle tool with layer toggler

### LINKED ISSUE #175: In layer viewers, switch from classroom icon to layer toggler.
Before we implemented the layer toggler with the checkboxes, we used a toolbar function (with the classroom icon) to turn the class layer on and off.

Everywhere we do this (I think it's in new Stages 4 & 5), that functionality should be moved to the layer toggler, so we can remove the classroom icon from the toolbar.

Guideline text will need to be edited to describe this slightly different (and better) way of handling the layers.
Screen Shot 2023-04-26 at 7 50 49 PM

Carifio24:
This PR looks to resolve #175 by replacing the class layer toggle tool with the layer toggler in stages 4 and 5. At the moment this only contains functionality updates - I haven't changed the guideline text. I'm also not sure exactly how we want the layout with this to be in stage 5 - once we decide that I'll change appropriately.

### DIFF
--- a/src/hubbleds/components/generic_state_components/stage_4/guideline_trends_data2.vue
+++ b/src/hubbleds/components/generic_state_components/stage_4/guideline_trends_data2.vue
@@ -11,7 +11,7 @@
     :state="state"
   >
     <template #before-next>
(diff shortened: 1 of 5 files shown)


johnarban:
I'll look this over and identify what guidelines need to be changed

patudom:
I am taking over this check from @johnarban

PR #205, opened 2023-05-01
update for hubble wavezoom tool

johnarban:
tiny PR which updates this to work with PR #198


Carifio24:
I mentioned this on Slack, but I think it makes more sense for the tool to check for the method existence, rather than add a stub method like this. The tool inherently already knows about the viewer, but doing it this way makes things more tightly coupled by having the viewer also need to know about the tool.

Carifio24:
I think we can get rid of the dotplot method too

PR #216, opened 2023-05-02
Hook up example galaxy to database

johnarban:
Example galaxy data is added to the database


patudom:
I think this is overwriting the 1st measurement with my 2nd measurement.
Screen Shot 2023-05-04 at 2 14 09 PM

PR #223, opened 2023-05-04
Fix more initialization issues

johnarban:
This fixes several sequencing issues. 
-  Makes sure the velocity and distance fill tools are on if the student returns to the appropriate marker. 
-  Stage 1 -> 2 transition is fixed. This is the issue raised in https://github.com/cosmicds/hubbleds/pull/191. The fix is to have the `next` button emit a "stage_complete" which calls a function. 
-  Fix team mode issue where fill value puts stage 3 into a weird state where it can't advance properly because `angsizes_total` isn't set properly. This should only affect team mode

### DIFF
--- a/src/hubbleds/stages/stage_3.py
+++ b/src/hubbleds/stages/stage_3.py
@@ -460,7 +460,7 @@ class StageTwo(HubbleStage):
 
         
     def _on_marker_update(self, old, new):
-        print_log(f"Marker update: {old} -> {new}")
+        #print_log(f"Marker update: {old} -> {new}")
         if not self.trigger_marker_update_cb:
             return
         markers = self.stage_state.markers
--- a/src/hubbleds/stages/stage_3.py
+++ b/src/hubbleds/stages/stage_3.py
@@ -239,7 +239,6 @@ class StageTwo(HubbleStage):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         
-        print('at beginning of init', self.stage_state.marker)
         
         dosdonts_slideshow = DosDontsSlideShow(self.stage_state.image_location_dosdonts)
         self.add_component(dosdonts_slideshow, label='py-dosdonts-slideshow')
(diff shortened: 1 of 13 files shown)


johnarban:
Issue: linedraw and linefit tools don't initialize properly.

patudom (commented):
enable_distance_tool turns on the little tape measure in the table, right? I don't think we need that to be on ever with the example galaxy measurements because those always get populated automatically by the calculation guidelines.

patudom (commented):
Same with [177942c]. I think there is some confusion about the sequencing since we split things up with the dot plot sequencing, but I don't think the enable_velocity_tool has to be true until the student has measured all their wavelengths, which would be on dop_cal6.

I think in stage_1.py, line 567 should be:
`        if self.stage_state.marker_reached("dop_cal6"):`

patudom:
Back to the [117942c], I would think that you would want line 376 to be
`        if self.stage_state.marker_reached('fil_rem1'):` 
because that needs to be active after all the thetas have been measured, but I just tried it with a local student, and it is not turning on for some reason.
Screen Shot 2023-05-07 at 7 24 10 PM

johnarban:
yeah, that is where we would want it. I am trying to see why that tool isn't activating (it calls the function). my guess is a similar reason as to why the best-fit tool and linedraw tools don't seem to activate which i don't know

johnarban:
possibly, it's because the "update_tool" method sends a message to the front-end, but the front end hasn't been created yet.

patudom:
Worst case scenario, if they come in at fil_rem1, we could back them up 1 marker? Then when they move forward, the tool is active?

patudom:
Everything else here looks good to me. We should apply your stage transition fix to all the other stages, but maybe that can go in a new PR.

johnarban:
So that's true, for all of these things, going back one and then forward triggers the action. We could either go back one for them, or tell them to go back one

patudom:
I think go back one for them. I doubt the will remember exactly where they were.

patudom (commented):
Oh, I think I found the button I needed to press to share these with you. Sorry I missed that earlier!

src/hubbleds/stages/stage_3.py
```diff
@@ -444,9 +451,16 @@ def set_x_lim(viewer):
...
+     #@print_function_name
+    def _update_state_from_measurements(self):
+        student_measurements = self.get_data(STUDENT_MEASUREMENTS_LABEL)
+        angsizes = student_measurements[ANGULAR_SIZE_COMPONENT]
+        self.stage_state.angsizes_total = angsizes[angsizes != None].size
+
         
     def _on_marker_update(self, old, new):
-        # print_log(f"Marker update: {old} -> {new}")
+        print_log(f"Marker update: {old} -> {new}")
```
johnarban, line 463:
```suggestion
        #print_log(f"Marker update: {old} -> {new}")
```

PR #239, opened 2023-05-07
All stages updates

patudom:
This fixes a bunch of lingering issues I noticed in testing. The commit descriptions should be self-explanatory.

I'll note here remaining issues I'm aware of, which I think are fine to run the pilot with.
- In the dos/donts slideshow, most of the images are outdated relative to how the viewer looks now. We can update those after the pilot.
- @johnarban correctly flagged that our question on guideline_lack_bias_reflect2.vue is confusing and would be hard for someone new to this to answer. I tried to update the question, but it might still be too hard, in which case we can drop it.


johnarban:
oopos - i never actually put this comment through. I think this is good. I think we could go without the `lac_bia2`. the previous question covers the general idea, and it's hard to be quantitativ.

PR #245, opened 2023-05-08
Read in example galaxy data

johnarban:
This is a PR to read in student example galaxy measurement data. While it works, I don't know really how it works. There is a function `update_example_galaxy_data` which mirrors one for the student data, but everything works with it commented out, so I am not sure by skipping that I am messing something up. 

It does a little bookkeeping to make sure that the read in data has two rows


src/hubbleds/story.py
```diff
@@ -400,7 +401,44 @@ def update_student_data(self, *args):
...
+            categorical = col in self.categorical_cols
+            ctype = CategoricalComponent if categorical else Component
+            comp = ctype(np.array(data))
+            new_data.update({comp: np.array(data)})
+        
+        example_galaxy_data = dc[EXAMPLE_GALAXY_MEASUREMENTS]
+        example_galaxy_data.update_values_from_data(new_data)
+        HubblesLaw.make_data_writeable(example_galaxy_data)
                 
+    def update_data_values(self, dc_name, values, index):
```
Carifio24, line 434: I don't see why we need this function - it looks exactly the same as the superclass implementation

Carifio24:
I'll take a proper look at this either later or early tomorrow morning.

src/hubbleds/story.py
johnarban: I thought that was from `Stage` in `phases.py`. Does `HubblesLaw` or `Story` inherit that somehow?

Carifio24: Oh oops, I thought this was in stage.py! Feel free to ignore, sorry

Carifio24:
I pushed a couple of commits to fix up a few issues that could pop up: components could get removed from the sample galaxy measurement data, the `.fits` extension not being removed in the sample galaxy names, etc. I don't 100% remember what the flow in this part is supposed to look like, but I think that things are working here now.

PR #257, opened 2023-05-16
Remove dotplot tutorial scrollbars

### LINKED ISSUE #236: Dot plot tutorial slideshow has gray vertical and horizontal scrollbars
I don't understand why the content doesn't just fit the available space. I looked at the .vue code for both the hubble slideshow and the dot plot slideshow and as far as I can tell, all the styling is identical.

hubble_exp_universe_slideshow.vue does not have gray scrollbars
Screen Shot 2023-05-06 at 4 19 53 PM

dotplot_tutorial_slideshow.vue does have gray scrollbars
Screen Shot 2023-05-06 at 4 24 14 PM

Carifio24:
This PR aims to fix #236 with some CSS modifications. There are two fixes here:

* First , I removed the `height: 70vh` from the `v-window`. The content of this slideshow doesn't need 70vh screen height (unless your screen is short).
* Next, I set the `v-row` (there's only one) to have `width: 100%`. I'm not completely sure, but my guess for why we need this here and not in, say, the angular slideshow video, is that we aren't using the `v-window-item` elements as the entire `v-window` content - they're side-by-side with the dotplot viewer, and I'm wondering if they expect to contain all of the window content. Since the `v-row` is the child of the `v-window`, setting its width to 100% fixes this. It also seemed to come with left and right margins of -12px, so I've removed those as well.

### DIFF
--- a/src/hubbleds/components/dotplot_tutorial_slideshow/dotplot_tutorial_slideshow.vue
+++ b/src/hubbleds/components/dotplot_tutorial_slideshow/dotplot_tutorial_slideshow.vue
@@ -47,7 +47,6 @@
 
         <v-window
           v-model="step"
-          style="height: 70vh;"
           class="overflow-auto"
         >
         <v-row>
@@ -225,8 +224,14 @@ export default {
 
 </script>
 
-<style>
+<style scoped>
 .no-transition {
   transition: none;
+}
+
+.row {
+  width: 100%;
+  margin-left: 0;
+  margin-right: 0;
 }
 </style>
--- a/src/hubbleds/components/dotplot_tutorial_slideshow/dotplot_tutorial_slideshow.vue
+++ b/src/hubbleds/components/dotplot_tutorial_slideshow/dotplot_tutorial_slideshow.vue
@@ -231,7 +231,7 @@ export default {
 
 .row {
   width: 100%;
-  margin-left: 0;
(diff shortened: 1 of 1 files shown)


src/hubbleds/components/dotplot_tutorial_slideshow/dotplot_tutorial_slideshow.vue
```diff
@@ -225,8 +224,14 @@ export default {
...
+<style scoped>
 .no-transition {
   transition: none;
 }
+
+.row {
+  width: 100%;
+  margin-left: 0;
+  margin-right: 0;
+}
```
johnarban, line 236: In looking on my browsers, this section doesn't seem necessary. Mine doesn't even use the margin values. But this does no harm. 

I think this is good to merge

johnarban (approved):

src/hubbleds/components/dotplot_tutorial_slideshow/dotplot_tutorial_slideshow.vue
Carifio24: At least for me, I need these (and I actually think they should be `!important`, now that I'm looking at it again). Without them, the row div actually hangs 12px outside of the dialog (and has 12px extra space on the right).

dotplot-scrollbars

PR #258, opened 2023-05-18
Guard velocity measurements

johnarban:
This provides a guard against students bad velocities by preventing `obswave_total` from increasing if there are bad velocities.


johnarban:
This now requires cosmicds/cosmicds#239 to work

Issue #262, opened 2023-06-02
Functionality to allow students to select confidence ranges from histogram and report % of data points in different ranges


patudom:
This plan has the guidelines at markers con_int2 and con_int2c in mind, but we will likely use at least some of this functionality elsewhere also (or instead).

We still need to think about how we will report all this information without it being overwhelming for the user, but if we build the functionality, we can figure out a way to lay it out on the screen.

- [x] Student clicks histogram bin that they think is the "most likely" value
- [ ] Selected bin is highlighted in a subset color
- [ ] The age of that bin is autofilled into the guideline
- [ ] The guideline reports what % of the data are in that bin
- [x] Student clicks on the high and low bounds of their confidence range
- [ ] The full selected range is a new subset of 1 color (including the "most likely" value bin)
- [ ] Guideline reports % of data that is within the whole confidence interval (all bins between low and high markers)

~~- [ ] The range of bins between the marker and the "most likely" bin are highlighted as the "low" subset, and the high values are a different-colored subset. (Or maybe we want those to be the same color - will have to try it and see if different colors looks too chaotic.)~~
~~- [ ] The low and high values are autofilled into the guideline~~
~~- [ ] The guideline reports % of data in the low subset, in the high subset, and totally enclosed between the low & high markers (including the "most likely" bin).~~

patudom:
We re-spec'ed how to handle the stats and percentage ranges. That functionality is implemented in https://github.com/cosmicds/cosmicds/pull/243 and #269 and #271. Closing this.

PR #263, opened 2023-06-05
create SingleBin and BinRangeSelect tools

johnarban:
This PR starts to address #262

It modifies the existing `hubble:binselect` and `hubble:singlebinselect` tools to optionally create highlighted ROIs (only one can be shown at a time), and exposes the user's selected values. The selections can be accesses at tool.selection which returns a dictionary.

A subset can also be returned. 

A possible mode of operation would be to turn ROI's off and then work with the subset created by the `subset_state` property.


patudom:
This should be generalized and moved to cosmicds. We might not necessarily use it immediately in hubbleds given the new sequencing, but it will be good functionality to have handy for the future.

patudom:
I made of this in #270 and am closing this for now.

PR #264, opened 2023-06-05
Class data and stage 4 wait handling

### LINKED ISSUE #261: Class size and stage 4 wait handling
- [ ] Increase stage 4 waiting condition to have a total of 10 classmates
- [ ] Allow app to know how many students are logged on from the class, so if there aren't as many as 10 students, the waiting screen will allow students to progress with fewer students.
- [ ] Once student gets to Stage 5, don't fetch any more class data (because it will be confusing if their histogram keeps changing while working on the confidence ranges).

Carifio24:
This PR looks to resolve #261 with the following changes:
* Bump the default minimum number of students that we need to have finished data collection to 10 for stage 4
  - If the class size is less than 10 (knowing this is dependent on https://github.com/cosmicds/cosmicds/pull/240), we set the required number to be the entire class
* Once a student has reached stage 5 (even if they go back), their class data no longer updates (we just skip this part of the `_on_timer` call)

### DIFF
--- a/src/hubbleds/stages/stage_5.py
+++ b/src/hubbleds/stages/stage_5.py
@@ -1,4 +1,5 @@
 from functools import partial
+from glue.core.subset import RangeSubsetState
 
 from numpy import where
 # from cosmicds.components.layer_toggle import LayerToggle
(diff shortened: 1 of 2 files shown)


patudom:
I created a batch of 5 fake students and put them in a new class to test this. When I ran this branch with cosmicds#240 locally, the "waiting" screen never appeared for any of my students even though others in the class weren't finished yet. I may have set things up incorrectly, so it would be helpful for @Carifio24 or @johnarban to take a look at this, or for us to look at it as a group.

patudom:
@Carifio24, the way we have things now, when is the last time a student fetches the data? Is it fetching every N (which I think is 30) seconds throughout Stage 4, or are there only specific times in Stage 4 where the data get updated?

Depending on how it's set up, we might want to do one more fetch at the very beginning of Stage 5, so the students are working with the maximal amount of data possible.

Also, can you help me expose in Stage 5 a state variable that gives the total number of students they have data for (including the user themselves)? I am going to add that number to text in one of the guidelines. (So it will say something like "1 student's age measurement is using 5 galaxies. Your class's age measurement is using `{ 5 * n_students}` galaxies."

Thanks!

Carifio24:
@patudom The data fetching happens every 30 seconds throughout stage 4. There's no particular trigger that fires when the student reaches stage 5 - the next time that the timed request would fire, if the student is at stage 5, it just doesn't. But we can change that to add in one last data grab when the student hits stage 5.

As for a state variable, that should be straightforward as well. I'll set something up to make the variable, and then you can put it in in a guideline.

PR #269, opened 2023-07-12
Add percentage and statistic selectors into Hubble

Carifio24:
This PR contains a basic implementation, built on top of #268, of adding the percentage and statistic selectors from https://github.com/cosmicds/cosmicds/pull/243 into the Hubble data story. I've only focused on the functionality here - the layout is bad, and I wasn't 100% sure where we wanted these.

To test this, go to stage 5, guideline `new1`.


patudom:
Everything here is in #275 (via #271), so I am closing it.

PR #271, opened 2023-07-13
Integrate stats and percentages into more viewers

patudom:
This is built off #269. I'm including this in case it's helpful to have the stats & percentage tools added to all the other histogram viewers in Stage 5.

Also, I added content to the HubbleDS.ipynb file that @johnarban shared with me. It's handy for being able to jump around to different markers and check the state without having to update the jump button in the team interface.


patudom:
Closing because everything here is in #275.

PR #272, opened 2023-07-13
Add descriptions to mean median mode

johnarban:
This builds on top of #271 to add descriptions to the statistics selectors and images. The images require [@cosmicds/cds-website#28](https://github.com/cosmicds/cds-website/pull/28). 

This also adds a missing guideline new8 (slide 239 on the mockup), but doesn't add a marker for it in the sequence.

PR #275, opened 2023-07-15
Update stage 5 guidelines with stats and percentage content

patudom:
This requires https://github.com/cosmicds/cosmicds/pull/243 to run (which has now been merged to main). It is built on #269, #271, and #272, so I am closing them.  

If you want to check this new sequence, it starts on 'mos_lik1' and goes through 'con_int3'.

Questions:

- In guideline_most_likely_value_reflect4.vue (originally guideline_new4.vue), this snippet of code is a conditional under the @next button:

```
      else {
        state.marker_forward = 1;
        state.trend_line_drawn = false;
        state.best_fit_clicked = false;
      }
```
@johnarban, is this important for this part of the sequencing, or is it leftover from an old guideline this was copied from?

- I renamed some of the free-response labels and tags (relative to what they were in the May implementation). I'm assuming this won't cause any problems within the database, or for code @johnarban is using to parse the state data?


patudom:
@Carifio24, I have 2 non-urgent questions about image_location handling.

- These lines of code refer to stage_three, which makes me think this is very old code that we aren't using anymore. Should we get rid of it, and can we just delete it without breaking other things? (The same  snippet of code appears in stage_4.py, so we should treat that the same way.) https://github.com/cosmicds/hubbleds/blob/034d1343f38a5dad5243772f29f57d1a991ed2ab/src/hubbleds/stages/stage_5.py#L811-L814
- The image_location is currently a state variable, I think because we copied this from other stages where we pass the location to other components. It doesn't look like that is necessary here, so would it be better to define the location within the main block of code and not make it a state variable?

johnarban:
yeah, @patudom that would've been a leftover. 

and we don't really need the location to be a state variable, especially since it is only for the mean, median, mode images. 

and yeah `_update_image_location` doesn't seem to get called anywhere is a relic. i think this was from back when we needed to change the image directory because voila was doing something weird with paths, but now everything comes from the cds-website

PR #280, opened 2023-07-18
General final updates before July pilot

patudom:
This builds on #275 and #279, so we could close those and just merge in this, or we can merge them PR by PR if it's easier to keep track of what's what that way.


patudom:
Closing because this is all in #282.

PR #285, opened 2023-07-19
Final checks for july implementation

patudom:
These are all pretty minor tidy-up edits. @johnarban, feel free to take a look if you have time. If you don't it should be fine to merge before morning.


johnarban:
these seem alright to me

PR #286, opened 2023-07-19
update selector styling

heywooddogwood:
remove unnecessary margin-ing, and a no-gutters tag.


heywooddogwood:
tied to https://github.com/cosmicds/cosmicds/pull/252 for cosmicds

Issue #289, opened 2023-07-21
Implement Angular Size measurement guard rail


patudom:
Choose highest & lowest possible measurements; if student's measurement is outside that range, make them remeasure.

PR #290, opened 2023-07-24
Guard angular size measurements

johnarban:
Implement #289

Adds a guard for the galaxy size in the distance tool. By default, the guard is not active. If activated, by default it will show the user a warning if the galaxy size is larger than 60 arcminutes or smaller than 6 arcseconds. This can be set using the `set_guard` method and can be turned on and off using the `(de)activate_guard` methods. 

Looking at the data, our smallest galaxy is 11 arcseconds, and the largest real measurements are about 12 arcminutes.

### DIFF
--- a/src/hubbleds/components/distance_tool/distance_tool.py
+++ b/src/hubbleds/components/distance_tool/distance_tool.py
@@ -97,8 +97,10 @@ class DistanceTool(v.VueTemplate):
         fov = self.widget.get_fov()
         widget_height = self._height_from_pixel_str(self.widget.layout.height)
         ang_size = Angle(((change["new"] / widget_height) * fov))
-        if not self.validate_angular_size(ang_size):
-            return None
+        valid = self.validate_angular_size(ang_size, change['new'] > 0)
+        # print(ang_size, change["new"], valid)
+        # if valid:
+        #     print('valid measurement')
         self.angular_size = ang_size
         self.measurement_count += 1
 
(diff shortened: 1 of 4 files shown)


patudom:
@johnarban, I've posted my updates to match the functionality of the velocity guard in Stage 1. This is the behavior I intend:

- If student makes good measurements, the number of galaxies to measure on marker rep_rem1 just counts down to 0 as normal.
- If student makes a bad measurement, their bad recorded value is entered into the table, and the guideline tells them they need to redo their measurement. The WWT view and the ruler are reset to their original state. They can't advance or click any other rows until they have fixed it.
- If student leaves the app and returns with a bad measurement in the table, their bad measurement row is highlighted and they can't advance past this guideline until they fix it.

One thing I've done differently in Stage 3 than Stage 1 is that in Stage 1, we don't serialize the `has_bad_velocities` or `bad_velocity_index` properties, but when the student leaves the app on the repeat remaining measurements guideline, the app still knows which row is bad and starts there. I'm assuming Stage 1 has a loop somewhere that checks all the measurements in the table. 

For Stage 3, I just serialized the state properties that track whether there is a bad measurement, and the outcome seems the same as what we did in Stage 1. Feel free to update that if you think it would be better to do it how we did in Stage 1.

If you think this is fine, please feel free to merge.

Regarding the mouse click UI, I will write up a new issue for @Carifio24 to look at.

johnarban:
This looks good!

Before merging, we should decide on the limits. 

- I think 15 or 30 arcminutes is a good upper limit. 
- The smallest galaxy is listed as 11 arcsec (only 5 x SDSS resolution). So perhaps 5 arcsec?

Below is the measured size distribution, and then I compared unique galaxies with their SDSS counterpart. 


(removing duplicates). This gives a measurement scatter of about 30 arcsec

Issue #293, opened 2023-08-04
distance_tool usability improvements


patudom:
In distance_tool if student clicks to start drawing their line, but then immediately releases the mouse without a drag, the measuring tool ends up in a "stuck" state, where there is no measuring line and no actions can be taken other than to stop the measurement and start over. 

Note: for easy testing, `student_id=4172` will open the app to the correct point (as long as we all use `update_db = False` when using this student.

https://github.com/cosmicds/hubbleds/assets/12750048/ffb93583-1656-4879-8e45-b01c455f9593

- [x] Proposed UI change (Jon):

- Click mouse down and up to set first end point of line.
- Move mouse to extend line (without having to hold down click).
- Click mouse down and up to set 2nd end point of line.
- If student puts 2nd end point on top of 1st end point (without moving mouse in between), keep tool in "drag mouse to extend line" mode.
- When user wants to edit their line, they click an end point to "pick it up," then they can move mouse to change the end point's location without having to hold down the mouse button.

- [ ] #304
- [ ] #305

patudom:
#296 addresses the distance tool usability. Closing this and will address #304 and #305 separately.

PR #296, opened 2023-08-08
Modify distance tool interaction to avoid stuck state

Carifio24:
This PR modifies the interaction of the distance-measuring tool to match the scheme described in #293. Now, the initial creation of the measuring line is done without dragging - after the tool has been activated, the first click places one endpoint of the line, and the line will follow the user's mouse until they make another click. If the user doesn't move between the first and second clicks, the second click is just ignored and the tool remains in the "drag to extend line" mode.

Note that the post-line creation interaction has not been modified - it's still the same grab-and-drag interaction.

### DIFF
--- a/src/hubbleds/components/distance_tool/distance_tool.vue
+++ b/src/hubbleds/components/distance_tool/distance_tool.vue
@@ -83,17 +83,17 @@
         {{ measuring ? 'Stop measuring' : 'Start measuring' }}
       </v-tooltip>
     </div>
-            <contrast-brightness-control 
-                inlineStyle="padding: 0.5em 0;"
-                :enabled=true  
-                :showContrast=false
-                :reset="reset_style"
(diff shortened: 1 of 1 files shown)


patudom:
In some circumstances we can't reproduce reliably, it is still possible to get the measuring tool into a state where the measured line disappears completely (presumably because we gave it a "zero" value), but the tool is still "active" and the user will be stuck. 

Two ways I've been able to make this happen:

- I double clicked on the same spot, or 
- I clicked for 1 endpoint, moved the mouse around, and brought the line back to the exact same endpoint and clicked again.
 
These actions don't consistently put the tool in the bad state. If we can't figure out how to reproduce or fix this, we'll make this its own issue and keep an eye on it.

Carifio24:
@patudom My latest commit changes the updating portion of the interaction to also not require dragging. I think it's where we want it to be now - let me know if you notice any issues.

PR #297, opened 2023-08-10
Use session for requests

Carifio24:
This is the companion to (and depends on) https://github.com/cosmicds/cosmicds/pull/261 - this PR updates the Hubble stages and components to make requests using a session with a proper authorization header for the backend server.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,9 @@ __pycache__/*
 */.ipynb_checkpoints/*
 .DS_Store
 
+# Environment variable deinitions
+.env
+
 # Project files
 .ropeproject
 .project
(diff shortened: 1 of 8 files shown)


Carifio24:
As far as using this locally goes:

- Place your `.env` file with your API key in the same directory as the notebook (i.e. `src/hubbleds/.env`)
- One way to load environment variables is to use the `dotenv` package, which you'll need to install (`pip install python-dotenv`). I've added a cell to the beginning of the HubbleDS notebook that will call that package's `load_dotenv` function to load in your environment variables

CC @patudom @johnarban @nmearl @heywooddogwood

PR #287, opened 2023-07-20
Fetch data at stage 5 and store class data size

Carifio24:
This PR adds an additional call to update the class data when the student first reaches stage 5. Once `max_stage_index` is at least 5, the repeated query for more class data is stopped (to keep the data constant as the student explores/answers questions about it), so this gives one last update.

Additionally, this adds a stage 5 state variable `class_data_size`, which is updated whenever the class data is updated.

PR #299, opened 2023-08-15
Update stage 5 guidelines 2

patudom:
This builds on #287 and includes a guideline telling students that their "class" only includes everyone who has finished their measurements at that point.

Also minor text edits.


patudom:
@heywooddogwood, you can ignore this PR for now. We are going to finalize the class data handling plans and then make sure the guideline text makes sense as a final step.

PR #302, opened 2023-08-23
Track stage progress and extract marker functionality

### LINKED ISSUE #298: Add state variables that track student's progress through a stage and story
This relates to the reporting of students' progress on the educator dashboard. Until now, we've been calculating progress by referencing a fixed list of markers in each stage. This means that if we add, remove, or rename guidelines/markers, that will break the progress calculation.

@johnarban proposes that we define the per stage progress and store it within the stage state, and store the over all story progress within the story state, so these values are always self-consistent for a particular class, even if we update guidelines between implementations.

One possible way to store stage progress: we know how long the marker list in each stage is, and we know what marker a student is on, so every time the student advances a marker to the furthest point they’ve arrived at, the `stage_progress` can be stored as something like `(index of furthest marker reached)/(N markers in stage)`, within `on_marker_update`. We won’t update this quantity if the student goes backwards or forwards to guidelines they’ve already been on.

Carifio24:
This PR aims to resolve #298 by adding a `progress` value in stage states that updates whenever the marker is changed. Additionally, the total marker count (`n_markers`) and the max marker index reached (`max_marker_index`) are stored as well (so `progress = max_marker_index / n_markers`).

Additionally, this PR extracts all of the marker-based functionality, which is currently duplicated across stage states, into a single `MarkerState` class. The relevant stage states (i.e. all of them except the slideshows) now inherit from this class.

@patudom @johnarban Let me know if there's anything else that we'd like to have with respect to this.

### DIFF
--- a/src/hubbleds/marker_state.py
+++ b/src/hubbleds/marker_state.py
@@ -65,4 +65,4 @@ class MarkerState(CDSState):
         if index is None:
             return
         self.max_marker_index = max(index, self.max_marker_index)
-        self.progress = self.max_marker_index / len(self.markers)
+        self.progress = self.max_marker_index / (len(self.markers) - 1)
--- a/src/hubbleds/marker_state.py
+++ b/src/hubbleds/marker_state.py
@@ -1,0 +1,58 @@
+from echo import CallbackProperty, add_callback, callback_property
+
+from cosmicds.phases import CDSState 
+
+
+class MarkerState(CDSState):
+    marker = CallbackProperty("")
+    markers = CallbackProperty([])
+    indices = CallbackProperty({})
+    advance_marker = CallbackProperty(True)
+    progress = CallbackProperty(0)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.marker_index = 0
+        self.marker = self.markers[0]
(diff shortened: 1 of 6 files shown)


src/hubbleds/marker_state.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def marker_index(self, marker):
+        return self.indices[marker]
+
+    def _on_marker_update(self, marker):
+        index = self.indices.get(marker, None)
+        if index is None:
+            return
+        self.max_marker_index = max(index, self.max_marker_index)
+        self.progress = self.max_marker_index / len(self.markers)
```
johnarban, line 68:
```suggestion
        self.progress = self.max_marker_index / (len(self.markers) - 1)
```
4566e01b863d7df854b49668ad617c3a5e389c24
We want the progress to be 0 for the first marker and 1 for the last marker, the same as the dashboard https://github.com/cosmicds/educator_dashboard/pull/31/commits/4566e01b863d7df854b49668ad617c3a5e389c24

Carifio24:
For the error above, student ID should be 4244.

patudom:
For 11/16 implementation, we will use the explicit marker list and have more time to check this out for the next implementation

patudom:
Closing this and will leave #298 open so we can ponder how to handle that in the solara version.

PR #325, opened 2023-11-09
UI fixes

### LINKED ISSUE #304: Update measuring tool instructions on `ang_siz4` to reflect new functionality (Pat).

patudom:
- Update text in various guidelines to improve clarity
- Add one new mc question to address issue of distribution's shape in addition to range
- Fix state issue in #311 
- Resolve #304


johnarban (approved):
There is only one minor wording suggestion, otherwise I think this is good to merge

PR #326, opened 2023-11-09
Update mc answer correct/neutral/wrong

### LINKED ISSUE #300: Update multiple choice correct, neutral, incorrect options
- [x] Go through entire story, make sure that the correct/neutral/incorrect assignments make sense. (We may be too generous in calling some incorrect things neutral). (Pat & John)
- [x] Make sure if there is an all-neutral question, that some response still registers on the educator dashboard. (Else 0 score, 0 tries, 0 choice is confusing). (John)

patudom:
@heywooddogwood and I went through these while @johnarban was away and I forgot to post this PR.

This will resolve #300.


johnarban (approved):
These make sense to me

PR #335, opened 2023-11-17
Set total distances during stage 3 initialization

Carifio24:
This PR updates Stage 3 to set `distances_total` stage state during the stage initialization. This should help fix the issue that we were seeing during the deployment today - at the least, it will make the issue go away on a refresh.


johnarban:
I tested this with a student (4363) and it solves the issue of not being able to click the tape measure after refreshing. So this should be good to merge. In testing I noticed an (un?)related error (#336)  coming out of stage 4 when I press the "runner" or "tape measure" buttons in stage 1 and 3

PR #347, opened 2024-01-25
fix the example galaxy gaurd

### LINKED ISSUE #340: Example galaxy distance measurement needs guardrails
Discovered this while testing #337. 

For the example galaxy measurements, we currently let students measure angular size however they like, but that can put them in a state where they can't advance.

If they measure an angular size that is basically 0, they can't advance past this guideline:
Screen Shot 2024-01-03 at 5 39 42 PM

And if they measure an angular size that is huge (leading to a distance that is basically 0) and the user tries to zoom in on the dot plot around the region where the distance is zero, the story crashes on this guideline. I'm assuming this is because the calculation of the corresponding angular size dot plot range is failing. (Because the displayed range of the distance measurement is set to match the measured range, this is only an issue if a student measures 0 for the distance, making that part of the distance range.)
Screen Shot 2024-01-03 at 5 34 57 PM

johnarban:
This PR fixes #340, preventing low measurements of the galaxy size and prevents the guideline from auto-advancing

### DIFF
--- a/src/hubbleds/components/distance_tool/distance_tool.py
+++ b/src/hubbleds/components/distance_tool/distance_tool.py
@@ -101,7 +101,8 @@ class DistanceTool(v.VueTemplate):
         # if valid:
         #     print('valid measurement')
         self.angular_size = ang_size
-        self.measurement_count += 1
+        if valid:
+            self.measurement_count += 1
 
     @observe('measuring')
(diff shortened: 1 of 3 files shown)


patudom:
Closing this and will leave #340 open so we can ponder how to handle that in the solara version.

PR #351, opened 2024-04-30
Stage 6

johnarban:
Implements #350 using stage 6


nmearl (commented):
This looks great. We'll merge after #350. One suggestion is that we should choose `.vue` file names to either be Vue-consistent (CamelCase.vue style), or python-consistent (snake_case.vue). I have no strong inclination (my IDE prefers the former by default), so @patudom and @johnarban, you're welcome to decide.

Carifio24:
@patudom I would agree with using snake_case. PascalCase is the Vue convention but I think having overall project consistency would be nice.

patudom:
@Carifio24, just to fill you in, we decided at our meeting today to stick with the vue convention of PascalCase for vue files. Nick pointed out that it can get hairy to keep track of whether you are making edits to the new solara version or the old voila version of a file when they are side-by-side, and having the PascalCase file names for the new files makes it easier to keep track.

PR #352, opened 2024-04-30
Update intro slideshow

patudom:
This PR transfers the remaining intro slide content from the legacy branch to the new solara-based main branch.

Issues:
- The WWT buttons on steps 4 & 5 don’t do anything yet because I don't know how to hook those up.
- `solara.Image` does not expose the alt text prop, so we had to switch to using `rv.Img` for 508 compliance. `rv.Img` has some bad interaction with the column/row layout that caused a blank div to spill over beyond the bounds of the image column and cover up some of the text in the column to the left (see screenshot). To get around this, per @johnarban's suggestion, I switched the relevant images to use `solara.HTML`, specifying the image content using ``.
Screenshot 2024-04-25 at 6 43 21 PM
- I used the `__init__.py` file for the intro slideshow as it originally was in @nmearl's example. I'm assuming that it will eventually have to go into its own `00-introduction` folder so we can build a separate distances slideshow in Stage 2, but I wasn't sure how to set that up.


nmearl (commented):
Re: using `__init__.py`, this is a consequence of Solara's auto-page pathing. The `__init__.py` file in the `pages` directory will always be the "landing page", and any nested directories (e.g. 00_introduction), would actually should up on the left as a page you can navigate to, but which is not the first page a user is shown when they visit the site.

Barring manually handling the pathing, I'd suggest just taking everything in the `def Page()` function and moving it to its own component.that is just imported. It's certainly not necessary, but would keep things clean. If you don't want to do this at the moment, @patudom, please feel free to open an issue so we don't forget.

patudom:
Thanks, @nmearl. I will take a look. If I don't figure this out quickly, I will open an issue.

nmearl:
Yep, this is perfect. Thanks!

PR #356, opened 2024-05-03
Stage 5

Carifio24:
This PR adds stage 5 to the Solara version of the Hubble story. This is very much still a draft - it only adds the first handful of guidelines and a viewer - but I wanted to submit this as a draft PR to share the method of setting up the glue viewers. What I had to do to get this to work (which I copied from `glue-solara`) is to set up the glue application and the viewers that we want inside of a `use_memo` call inside of the page setup. Without this, attempting to create a viewer was consistently giving me issues with widgets closing, as a viewer actually consists of several widgets together (e.g. the figure widget, the layer options widget, and the viewer options widget).

### DIFF
--- a/src/hubbleds/components/plotly_viewer/plotly_viewer.py
+++ b/src/hubbleds/components/plotly_viewer/plotly_viewer.py
@@ -1,5 +1,0 @@
-import solara
-
-@solara.component
-def PlotlyViewer(
-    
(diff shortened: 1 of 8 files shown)


Carifio24:
@nmearl Let me know if you have any thoughts on the Solara-based stuff here (which should mostly just consist of the page itself and the ID slider component)

nmearl (approved):
Looks great to me!

PR #364, opened 2024-05-08
Stage 6

johnarban:
This PR implements the stage 6 guidelines and professional data viewer. It requires [cosmicds#281](https://github.com/cosmicds/cosmicds/pull/281) for the LayerToggle and LineFitTool. _line fit tool is not implemented for Plotly yet_.


src/hubbleds/pages/06-prodata/__init__.py
```diff
@@ -0,0 +1,208 @@
...
+        Can advance: {component_state.can_transition(next=True)}
+
+        mc-scoring: {mc_scoring}
+
+        """
+    )
+    
+    
+    with solara.Row():
+        with solara.Column():
```
nmearl, line 91: I found that using `solara.Row` and `solara.Column` resulted in incorrect column widths (they're equal sized instead of 1:2). Is this turning out okay for you? If not, I'd suggest using `rv.Row` and `rv.Col` as in the Spectra & Velocity page.

nmearl (changes requested):
This looks good overall. Just a few comments. Also, I would say that we should either keep the `HubbleView*` stuff separated into their own files (like the `hubble_dotplot`), or merged all in the `viewer` file, for consistency.

src/hubbleds/pages/06-prodata/__init__.py
johnarban: I had not paid attention to the styling. I will update it to use the rv.Col with the appropriate cols=4 and 8 widths

johnarban (commented):
fixed the layout per nick's comments

src/hubbleds/pages/06-prodata/__init__.py
patudom: I used `solara.ColumnsResponsive` as in [here](https://github.com/patudom/hubbleds/blob/bd4ca86818de818d679ef505b3ba672ae2b23140/src/hubbleds/pages/03-distance-measurements/__init__.py#L99), which ports over how we handled this in the voila version. I found that solara didn't really care if I included Rows or not, so I think I left those out in some places and it behaved correctly.

nmearl (approved):

PR #371, opened 2024-05-17
Modify guidelines to use method in MathJax intersection observers

Carifio24:
@nmearl and I discussed this and decided that using a method was clearer than making the window-level `MathJax` object be a computed value. This PR updates existing guidelines to use that method for consistency.


nmearl (approved):
Nice. Thanks, @Carifio24!

PR #378, opened 2024-05-24
Update stage 5 guidelines solara

patudom:
This adds the remaining Stage 5 guidelines and lays them out on the page in the same row/column configuration from the voila version. "Team-interface"-pink cards have been inserted as placeholders for the viewers and other components that still need to be hooked up.

I added a `MarkerBase` class method to make it easier to specify marker ranges where content should appear. Per a conversation with @Carifio24 , he will update that to streamline it further and convert all the Marker transition functions that we currently have in each stage's `component_state.py` file into a base class that gets imported.

We had various placeholder notes that either started with "FIX" or "TODO" and I decided to make them all "TODO" to make it easier to find everything to deal with later.

Note, for some reason, solara ignores `<strong>` in all our view files, so I changed them all to `<b>`.


nmearl:
@Carifio24 While this is definitely a way to handle this, given that we redefine `Marker` in each stage, an alternative might be to simple use the marker value in the component state instead of the marker instance. That is, instead of

```python
current_step: Reactive[Marker] = dataclasses.field(
        default=Reactive(Marker.mee_gui1)
    )
```
we can use
```python
current_step: Reactive[int] = dataclasses.field(
        default=Reactive(Marker.mee_gui1.value)
    )
```
and update the marker manipulation accordingly. That said, I have no issue with a factory function. Likewise, it's also possible to update the marker manipulation behavior to be `classmethods` instead of `staticmethods`; e.g.

```python
    @classmethod
    def next(cls, step):
        return cls(step.value + 1)
```
to ensure that it's always referencing its own class. However, this solution is more than fine currently.

Carifio24:
@nmearl That's an interesting idea to change the type of `current_step` - I guess then we'd have to define `current_step` in each component state, or do something like.
```python
current_step: Reactive[int] = dataclasses.field(default=0)
```

(since `0` will always be the default, I'd imagine)

Unless I'm missing something that you're saying, I think we would still have issues with some of the standard component state methods, though - `can_transition`, `transition_previous`, `transition_next`, etc all require a reference to the particular marker class since they involve calling class methods. The need to have a runtime reference to the type for those operations is what ultimately pushed me in the factory function direction.

nmearl:
> I think we would still have issues with some of the standard component state methods, though - can_transition, transition_previous, transition_next

Yeah, for sure. They're `staticmethods` right now, which requires a reference to `Marker`, but if we convert them to `classmethods`, then we use it as above:

> Likewise, it's also possible to update the marker manipulation behavior to be `classmethods` instead of `staticmethods`; e.g.
> 
> ```python
>     @classmethod
>     def next(cls, step):
>         return cls(step.value + 1)
> ```

Carifio24:
@nmearl Ah, okay, I think I get what you're saying. If `next` and `previous` are class methods in `MarkerBase` (which they actually already are), then we can get e.g. the next marker as 


```python

step = self.current_step.value.next(self.current_step.value)

```

without needing a reference to the marker type. Not sure why I didn't think of that to begin with!

nmearl:
This seems to work well for me on a cursory check. Feel free to merge when ready!

PR #379, opened 2024-05-28
Make MCScore Reactive

johnarban:
This PR makes MCScore reactive. I also create a function to serialize the mc_score for use in `state_view` from ScaffoldAlert. It also makes the gate in stage 5 work properly


Carifio24:
Just curious, what was the reason for reordering the marker base classes and removing the `MarkerBase` metaclass?

johnarban:
Ahh, @Carifio24, I would get the following errors - 

```
File ~/anaconda3/lib/python3.9/enum.py:611, in EnumMeta._get_mixins_(class_name, bases)
    609 first_enum = bases[-1]
    610 if not issubclass(first_enum, Enum):
--> 611     raise TypeError("new enumerations should be created as "
    612             "`EnumName([mixin_type, ...] [data_type,] enum_type)`")
    613 member_type = _find_data_type(bases) or object
    614 if first_enum._member_names_:

TypeError: new enumerations should be created as `EnumName([mixin_type, ...] [data_type,] enum_type)`
```
This would occur when Marker was declared using `class Marker(enum.Enum, MarkerBase)`


And the other error was,
```
ERROR:    Component react.component(hubbleds.pages.06-prodata.Page) raised exception AttributeError("'int' object has no attribute '_member_type_'")
Traceback (most recent call last):
  File "/Users/jal194/anaconda3/lib/python3.9/site-packages/reacton/core.py", line 1707, in _render
    root_element = el.component.f(*el.args, **el.kwargs)
  File "/Users/jal194/github/cosmicds_dev/hubbleds/src/hubbleds/pages/06-prodata/__init__.py", line 78, in Page
    f"""
  File "/Users/jal194/anaconda3/lib/python3.9/enum.py", line 772, in __format__
    return cls.__format__(val, format_spec)
  File "/Users/jal194/anaconda3/lib/python3.9/enum.py", line 765, in __format__
    if self._member_type_ is object or str_overridden:
AttributeError: 'int' object has no attribute '_member_type_'
```
This error occurs when you do something like `solara.Markdown(f"{Marker.pro_dat0}")`. I don't really understand metaclasses, but I think the way MarkerBase was defined left something out. The first error happens whether or not I included the metaclass

Carifio24:
@johnarban Must be an issue between versions of the Python `enum` library - from the traceback it looks like you're on 3.9; I was using 3.11 locally when I set that up (and the referenced line in `enum.py` is something different on my end).

I was mostly just curious - if this works fine then we should use what's more generally compatible. The metaclass was mostly to ensure that any inheriting `Marker` classes were actually enums, but I don't think that really matters.

johnarban:
So I have reverted the changes in MarkerBase and Marker base classes and fixed a couple bugs that seemed to creep in when I updated my python version to 3.11. I have merge `main` into this branch, and fixed stage 5 to use the new MCScore.

johnarban:
This has been updated to fix all the guidelines with mc-radiogroup to use the new format and to allow advancing via the gates. There are few places where the mc-radiogroup is in a sub-component with, those can be done in a separate PR

patudom:
Does anyone else get a flicker in the "explainer text" that appears in the box when going from wrong/neutral to right answers? I'm seeing this consistently for some questions but not others. This is Stage 4, `rel_vel1`. The questions in Stage 5 don't seem to have this flicker.

https://github.com/cosmicds/hubbleds/assets/12750048/63f71ecc-84e2-4b46-8ad0-b0492a2e6da2

patudom:
The very last mc question in Stage 6 `pro_dat9` is behaving especially strangely. I have to press all the radio buttons twice for them to respond to anything, and then the "next" button never appears. (It gets cut off in the video, so you can't see it).


https://github.com/cosmicds/hubbleds/assets/12750048/b53477bf-fca8-427a-9f2b-6980014d1c83

PR #386, opened 2024-06-05
Add dependencies

Carifio24:
This is the HubbleDS companion to https://github.com/cosmicds/cosmicds/pull/287.


nmearl:
So currently the python build ecosystem is still a mishmash of competing systems. Currently, the template we use for the package recommends sticking with `setup.cfg` over `pyproject.toml`. It does not matter for most things, but there is already an `install_requires` in `setup.cfg`, along with a lot of our package-specific install setup. I would recommend just sticking with `setup.cfg` for now, as it is still the most compatible. Ultimately, I don't think it really matters for this case, but eventually we'll want to try just ditching `setup.cfg` and move everything over to to `pyproject.toml`.

Carifio24:
> Currently, the template we use for the package recommends sticking with setup.cfg over pyproject.toml

Ah, I didn't realize that. In that case your suggestion sounds great to me - stick with `setup.cfg` for now and migrate to `pyproject.toml` at some point in the future. I've modified these PRs to list the dependencies in `setup.cfg`, and remove `pyproject.toml` for the moment to keep things simpler.

nmearl (approved):

PR #400, opened 2024-06-08
Additional dependencies and environment file/script

### LINKED ISSUE #380: Create environment file for development
We should create a YAML file that defines a consistent development environment (and can then be used to recreate that environment via conda).

Carifio24:
This PR updates our dependencies with a couple of additional libraries (`authlib` and `itsdangerous`). Solara complained when these were missing, but didn't seem to pull them in itself, so I've added them to our dependency list here.

Also, this PR looks to resolve #380 by adding an environment file, as well as a script to update it based on the current environment. The basic idea here is:
* We only really care about the Python packages that we're installing (especially as someone who switches OSes somewhat frequently, I don't want to run into issues between base conda dependencies across machines). So everything other than pip and what it installs are pruned out.
* Make sure that the environment file installs `ipywwt` from git
* Remove any reference to an installed `cosmicds` or `hubbleds`, since this is supposed to be a development environment for this package. This point I was a bit unsure on and I'm happy to go either way.


Carifio24:
As far as testing this goes, try creating an environment with

```bash
conda env create --name <env-name> --file environment.yml
```

then install `cosmicds` and `hubbleds` from source, and check whether the app works

PR #407, opened 2024-06-11
Line draw tool

### LINKED ISSUE #391: Implement Stage 4 Line Draw tool on hubble graphs
Stage 4 hubble viewers will all be implemented in plotly

(Comment from closed #382: Investigate whether the logic we used in bqplot can be supported by plotly or whether we should go with a canvas approach (cribbed from the WWT angular size measuring tool), we we are thinking that we can do this without a canvas approach. )

Carifio24:
This PR resolves #391 by implementing the line drawing functionality in Plotly. This requires https://github.com/cosmicds/cosmicds/pull/289 as the bulk of the work here is done using JavaScript. This PR also contains a quick example of passing in some sample data to this viewer as Plotly traces; this still needs to be updated to be the actual data that we want at this point in the story. I just wanted to post this as a PR so that it can be experimented with and we can decide whether we like the UX and interactions.

Here's a quick demo below of this in action. You can only interact with the line when the drawing "tool" is active. The flow is similar to the old bqplot-based tool - first draw a line, and then you can click on the endpoint to make that line movable again.

https://github.com/cosmicds/hubbleds/assets/14281631/b6d4b3ac-88b4-40da-bbcd-28fa20bec743

### DIFF
--- a/src/hubbleds/components/line_draw_viewer/LineDrawPlot.vue
+++ b/src/hubbleds/components/line_draw_viewer/LineDrawPlot.vue
@@ -69,7 +69,7 @@ export default {
       element: null,
       lineDrawn: false,
       mouseDown: false,
-      movingLine: true,
+      movingLine: false,
       lastEndpoint: null,
       plotDataCount: 0,
       lineTraceIndex: 0,
@@ -105,7 +105,6 @@ export default {
         this.movingLine = false;
         this.drawEndpoint(event);
         this.lineDrawn = true;
-        this.setCursor("move");
         if (this.line_drawn) {
           this.line_drawn();
         }
@@ -132,7 +131,8 @@ export default {
     plotlyUnhoverHandler(event) {
       if (event.points[0].curveNumber === this.endpointTraceIndex) {
         this.hoveringEndpoint = false;
-        this.setCursor("crosshair");
+        const cursor = this.movingLine ? "grabbing" : "crosshair";
+        this.setCursor(cursor);
       }
     },
     setCursor(type) {
@@ -202,9 +202,14 @@ export default {
       );
     },
     active(value) {
+      this.movingLine = value && this.lastEndpoint === null;
       Plotly.update(this.chart.uuid, { visible: true }, {}, [this.lineTraceIndex]);
       this.setupMouseHandlers(value);
       this.setupPlotlyHandlers(value);
+    },
+    movingLine(value) {
+      const cursor = value ? "grabbing" : "move";
+      this.setCursor(cursor);
     }
   }
 }
--- a/src/hubbleds/components/line_draw_viewer/LineDrawPlot.vue
+++ b/src/hubbleds/components/line_draw_viewer/LineDrawPlot.vue
@@ -26,6 +26,7 @@ export default {
       showgrid: false,
       showticklabels: true,
       autorange: false,
+      automargin: true,
     };
     const xaxis = { ...baseAxis, range: [0, 1] };
     const yaxis = { ...baseAxis, range: [0, 1] };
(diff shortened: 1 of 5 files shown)


patudom:
This is looking great, @Carifio24. Thanks!! The functionality for drawing the line and picking up the end to move it works well.

We should think about what we want the button to do if the student clicks it again. Currently, that seems to "fix" the line in place and it is no longer editable or removable, which we will need to change.

https://github.com/cosmicds/hubbleds/assets/12750048/9a474a5e-18fd-4ada-9be6-54b43284e492

Maybe the button icon can change to an eraser when the line is drawn, so it's clear if you click it again, the line goes away, then you reactivate the draw tool by pressing the pencil button and then draw a new line.

Carifio24:
@patudom My idea had been that if the student re-activates the button, they can then edit the line in the same way (clicking the endpoint to make the line editable). It was a bug that was preventing that from working but the most recent commit fixes that.

patudom:
@Carifio24, editing the line now seems sluggish (but I don't think it's because of the change you just made - I just tried reverting it and it's doing the same thing). I am in the new environment that I created using #400 - is it possible something about that has triggered these issues?

It takes a while for the line to snap to where I moved the cursor. 
https://github.com/cosmicds/hubbleds/assets/12750048/e542f1d6-be01-4f60-a04c-be146b9323f4

And at one point, I somehow separated the end dot from the line itself. (i.e., I could move the dot, but the line didn't follow).
Screenshot 2024-06-12 at 12 24 36 PM

patudom:
(I just tried running this in my original solara environment and it's still being sluggish. It seems a little concerning that this behaved fine on my computer earlier today and is now slow..)

Carifio24:
@patudom That's strange - the updated environment really isn't very different from the old one, and all of this is happening on the JavaScript side so the updated Python packages shouldn't have any effect. When the line movement is slow, how does the rest of the app run?

I'll see if I can reproduce the line/dot separation issue.

patudom:
Oh, I figured out one issue that was my bad. I was checking the UI in the voila version, and to edit the line there, you click the end and hold the mouse down while you're dragging. I inadvertently started doing the same thing here, which is not the correct way to move the line in this UI. Here, you click the end and move the mouse without holding it down. This makes sense because it matches the line draw behavior in the distance tool. But sometimes, when I click the end of the line, it doesn't register that I've clicked, so it seems like nothing is happening. I'm not sure under what circumstances the click registers or doesn't.

https://github.com/cosmicds/hubbleds/assets/12750048/ae5de980-f07b-45ff-9c22-ca16adee3050

patudom:
> @patudom That's strange - the updated environment really isn't very different from the old one, and all of this is happening on the JavaScript side so the updated Python packages shouldn't have any effect. When the line movement is slow, how does the rest of the app run?
> 
> I'll see if I can reproduce the line/dot separation issue.

Just to clarify, the environment was NOT the issue. It was doing the same thing in my original solara environment.

patudom:
- [ ] When the user is editing the line endpoint, we should change the mouse icon to a Mickey Mouse hand to match the distance tool

Carifio24:
> and to edit the line there, you click the end and hold the mouse down while you're dragging

Ah okay, I thought that we had changed the Voila version to not use dragging at some point. I like the cursor hand idea, I'll  add that in.

patudom:
> > and to edit the line there, you click the end and hold the mouse down while you're dragging
> 
> Ah okay, I thought that we had changed the Voila version to not use dragging at some point. I like the cursor hand idea, I'll add that in.

Thanks, Jon! And just to be clear, I think the way you have it here is what we want. (Maybe we meant to change the behavior in the voila version and never got around to it. 🤷‍♀️ )

Carifio24:
Just added a commit that adds in the grabbing cursor.

PR #408, opened 2024-06-11
implement free-response questions

johnarban:
This implements the free response component (in combination with https://github.com/cosmicds/cosmicds/pull/290). This uses `get_free_response` to initialize and pass values to the free-response and `fr_callback` to handle events. 

There are two issues:

**Flickering** One issue, is that it flickers each time we remove focus from the text element. The reason is that the entered value is sent to the backend when the text element is defocused. This triggers a render. This is a problem if the user clicks "next" right after entering their value, as the "blur" event occurs first, followed by re-rendering, and so the "next" is not completed. Then "next" must be pressed again. I am not quite sure the best way around this. 

**Error state doesn't persist** after text box loses focus. The error state (check `pro_dat8`) goes away when not focused and then is inconsistent if you enter/leave it

### DIFF
--- a/src/hubbleds/components/uncertainty_slideshow/UncertaintySlideshow.vue
+++ b/src/hubbleds/components/uncertainty_slideshow/UncertaintySlideshow.vue
@@ -136,7 +136,7 @@
                   <p class="StudentResponses">
                     TODO: add free response box
                 </p>
-                  <!-- <free-response
+                  <free-response
                     outlined
                     auto-grow
                     rows="2"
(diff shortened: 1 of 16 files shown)


patudom:
Merging this now, so we can' keep building on it, but there are still a couple issues we have to come back and fix:

- [ ] @johnarban experiences flickering if the user clicks next quickly. This may or may not be a different manifestation of the same problem, but I have to click "next" twice for every input in order to move forward. (I don't see a flicker).
- [ ] We need fix the tabbing behavior for guidelines with multiple entry boxes. For some reason, tabbing does not take you to the next input box, but it takes you everywhere else on the page first. Try explicitly setting tab indices.
- [ ] We were having issues with the uncertainty slideshow where it wasn't being correctly initialized with the free response values. John has a workaround where responses are allowed to be initialized as undefined [here](https://github.com/cosmicds/hubbleds/pull/408/commits/d4a0d0b8f141fcb7622c363c67e478f6204d0cc1), but there may be a deeper underlying issue where things are firing out of order. (The initialization issue was only breaking things on my computer, not John's).

PR #411, opened 2024-06-13
Fix free response reactives

### LINKED ISSUE #409: Lingering free-response issues
Issues remaining from #408:

- [x] @johnarban experiences flickering if the user clicks next quickly. This may or may not be a different manifestation of the same problem, but I have to click "next" twice for every input in order to move forward. (I don't see a flicker).
- [ ] We need fix the tabbing behavior for guidelines with multiple entry boxes. For some reason, tabbing does not take you to the next input box, but it takes you everywhere else on the page first. Try explicitly setting tab indices.
- [ ] We were having issues with the uncertainty slideshow where it wasn't being correctly initialized with the free response values. John has a workaround where responses are allowed to be initialized as undefined [here](https://github.com/cosmicds/hubbleds/pull/408/commits/d4a0d0b8f141fcb7622c363c67e478f6204d0cc1), but there may be a deeper underlying issue where things are firing out of order. (The initialization issue was only breaking things on my computer, not John's).

_Originally posted by @patudom in https://github.com/cosmicds/hubbleds/issues/408#issuecomment-2163895298_

johnarban:
The cause of the need to double-click the next button (#409) to advance was due to `FreeResponse` having reactive variables. The response does not have to be reactive. All error state handling is done on the front end, and the  Free Response dataclass just stores the answers. This should fix #409

### DIFF
--- a/src/hubbleds/free_response.py
+++ b/src/hubbleds/free_response.py
@@ -11,8 +11,8 @@ from typing import Union, Dict
 @dataclasses.dataclass
 class FreeResponse:
     tag: str = dataclasses.field(init=True)
-    _response: str = dataclasses.field(default_factory = lambda: "")
-    _initialized: bool = dataclasses.field(default_factory = lambda: True)
+    _response: str = dataclasses.field(default = "")
+    _initialized: bool = dataclasses.field(default = True)
     
     def update(self, response: str = ''):
         # self._response.set(response)
--- a/src/hubbleds/free_response.py
+++ b/src/hubbleds/free_response.py
@@ -36,7 +36,7 @@ class FreeResponse:
     
 @dataclasses.dataclass
 class FreeResponseDict:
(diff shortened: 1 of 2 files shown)


src/hubbleds/free_response.py
```diff
@@ -11,27 +11,27 @@
 @dataclasses.dataclass
 class FreeResponse:
     tag: str = dataclasses.field(init=True)
-    _response: Reactive[str] = dataclasses.field(default_factory = lambda: Reactive(""))
-    _initialized: Reactive[bool] = dataclasses.field(default_factory = lambda: Reactive(True))
+    _response: str = dataclasses.field(default_factory = lambda: "")
```
nmearl, line 14: Is there a reason that these are factories and not just regular fields?

johnarban: I suppose not anymore. Previously (when testing `MCScore`), I found the each copy of the `MCScore` dataclass would have the same `Reactive` variable in it. But since these aren't reactive, I will change them back

PR #417, opened 2024-06-13
FreeResponse does not need any reactivity

johnarban:
After some discussion with @patudom, this PR removes the reactivity from `FreeResponseDict` and `FreeResponse`, since we only ever store or retrieve the responses. Callbacks from the appropriate guidelines update the values and they get called when needed. 

There are `default_factory`'s here because python complained that you can't have mutable objects as defaults. I am not sure if this makes things better or worse re:saving the state

### DIFF
--- a/src/hubbleds/free_response.py
+++ b/src/hubbleds/free_response.py
@@ -31,34 +31,34 @@ class FreeResponse:
     
     @computed_property
     def completed(self):
-        return self._response is not ""
+        return self._response != ""
     
     
 @dataclasses.dataclass
 class FreeResponseDict:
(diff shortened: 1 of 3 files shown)


patudom:
Thanks, @johnarban - this works for me. @nmearl, please let us know if it's ok with you to merge this.

patudom (approved):

nmearl:
I think this is definitely in the right direction! I'm mucking around with the new API @Carifio24 put together, and fixing the globally reactive state issues, so I can't test right now. But on its face, this looks good -- feel free to merge, and I can consult with @johnarban as a follow up if there are issues!

PR #418, opened 2024-06-13
Connect free response inputs to guidelines

patudom:
This PR fixes the guidelines that display the free responses the user input in earlier guidelines. We are now reading the responses directly from `LOCAL_STATE.free_responses` so we don't need to include duplicate reactive variables in the component_state.

I also fixed a minor layout issue in Stage 5 where the uncertainty slideshow button was disappearing when it wasn't supposed to.


johnarban (approved):
added in mathjax support with 61107a15468d70e421719357791fa25b004ff891

PR #421, opened 2024-06-15
Add statistics and percentage selector components to stage 5

Carifio24:
This PR adds the statistics and selector components from https://github.com/cosmicds/cosmicds/pull/293 to stage 5 of the story. For now, these components are connected to a histogram viewer with the test data that was already in there (it conveniently has multiple modes, which was nice for testing).


patudom:
I'm not sure if this comment makes more sense here or at https://github.com/cosmicds/cosmicds/pull/293, but in addition to the comment at the other repo about the histogram bars not being centered, the switch settings need to be "remembered" if a user goes back to a guideline. In this case, I turned on "mode," moved ahead through some guidelines and came back to this guideline, but all the switches are now in the "off" position. There is now no way to wipe the mode from the viewer. (Switching it on and back off does not erase it, so it is stuck).

Screenshot 2024-06-17 at 10 41 42 AM

Carifio24:
@patudom I'll add the currently selected options for both of these (percentage and statistic) to the stage state, so that we can correctly set up the components each time they're recreated.

Issue #423, opened 2024-06-17
Using theme colors in solara


patudom:
Is there a way to specify theme colors (primary, secondary, etc), inside solara `style` specifications, like here? https://github.com/johnarban/hubbleds/blob/051c87c15d12ee78359dd844c5ab1a0499a5a71f/src/hubbleds/components/grid_viewer/grid_viewer.py#L12

If we ever decided to change the theme colors, this usage will be very messy to update.

patudom:
Per @johnarban's suggestion, we now import our theme colors into the css [here](https://github.com/cosmicds/hubbleds/pull/425/commits/4c83f1180a302d8ec4e6cc8197e5ca93516a181f), but we still need to wire up functionality that allows the css colors to respond to light/dark theme changes.

PR #425, opened 2024-06-17
Stage 6 viewer layout

patudom:
A couple small changes to update the layout of the layer toggle + viewer in stage 6.


patudom:
This requires https://github.com/cosmicds/cosmicds/pull/295.

In addition to updating the Stage 6 viewer layout, this now also:
- Starts adding theme colors to our css per #423 (but note that our css does not yet respond to light/dark theme changes.)
- Removes the grid_viewer component & moves it to [cosmicds](https://github.com/cosmicds/cosmicds/pull/295) since it is a generic functionality not specific to hubble.
- Adds viewer titles to the Stage 5 viewers

patudom:
Note, for the guideline `pro_dat0`, the layer toggle component appears as a box with ~0 height. This will be fixed once we have the class data wired up to the viewer.

PR #432, opened 2024-06-21
Improve light/dark theme functionality

patudom:
This works toward improving the light/dark theme functionality as noted in #423. 

Many things are working better now, but there are still some flaky cases where I couldn't figure out how to target what we needed. Where this happens, I try to favor the dark theme functionality.

- Colors in custom.css & theme.js are now consistent and should match what's at https://github.com/cosmicds/educator_dashboard (where I had made earlier changes from the voila-hubbleds to improve contrasts).
- Where we specify theme colors by hex codes, I've changed those back to the theme colors.
- Straight vue files seem to handle the light/dark themes automatically, but things that were specified in python/solara often ignored the theme changes, so I copied the css used for the [solara webpage](https://github.com/widgetti/solara/blob/d258da07ff2ce698553e111312b2e2bc8763ea00/solara/website/assets/custom.css#L266) to make sure it updated correctly.
- One pesky issue was the border around the MathJax equation `.legend`s, as [here](https://github.com/patudom/hubbleds/blob/02ddb24447fee2e23c54f6553da21eb600edff63/src/hubbleds/pages/01-spectra-%26-velocity/guidelines/GuidelineDopplerCalc2.vue#L111). When I tried moving this to `custom.css`, it was completely ignored, and I couldn't split this into dark/light mode cases. (I think this is a problem that we never solved in the voila version either).
- Some of the toolbar buttons are gray in light mode instead of staying white.

### DIFF
--- a/src/hubbleds/assets/custom.css
+++ b/src/hubbleds/assets/custom.css
@@ -1,68 +1,78 @@
 :root {
-    /* define light material desing them */
-    --primary-light: #1565C0 ;
-    --secondary-light: #006064 ;
-    --accent-light: #FFD740 ;
-    --error-light: #F06292 ;
-    --info-light: #FF7043 ;
-    --success-light: #388E3C  ;
-    --warning-light: #DD2C00 ;
+    --primary-dark: #0D47A1;
+    --secondary-dark: #00838F;
+    --accent-dark: #FFC400;
+    --error-dark: #D81B60;
+    --info-dark: #BF360C;
+    --success-dark: #00E676;
+    --warning-dark: #DD2C00;
     
-    /* define dark material desing them */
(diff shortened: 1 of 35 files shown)


johnarban (approved):
I have added a couple commits to fix the issue with the MathJax legend border https://github.com/cosmicds/hubbleds/pull/432/commits/3b174d78130677160bd48e8291cbab473e7dc635 and the button color 7b7717ff9c597a502d33c8fde2cddead9c0063b6, and in a commit on [cosmicds/340be9c](https://github.com/cosmicds/cosmicds/pull/297/commits/340be9c09707605424a6ac39b640811a92f83a59) for the toolbar figure title

PR #433, opened 2024-06-23
Add in stage 3 dotplot viewers

johnarban:
This hooks up the dotplot viewers to real data in the stages 1 and 3. 

In stage 3, it is setup to show the students data on the graph. Sequencing has not been implemented, but should be pretty simple. 

Currently I am not using the `DotplotViewer` @nmearl made, as while I was making this it did not allow data to be added (I have fixed this, but it currently only takes in a single data set (I suppose I can fix this too). 

A major outstanding issue is adding the red line. I can't seem to hook into the `viewer.figure.selection_layer.on_click `callback. However, it seems like we ought to be able to use the click event the same way[ glue uses the selection event](https://github.com/glue-viz/glue-plotly/blob/34e231661e53dc49571d85c0be98437410692f78/glue_plotly/viewers/common/tools.py#L36). However in my attempt, I couldn't get the click callback?? 

In the spectrum viewer, @nmearl get's around this using the plotly figure object directly. 

I am not quite sure how to go about this, perhaps @Carifio24 or @nmearl have ideas. I didn't see any built-in glue-plotly tools which used the single click (I think the nearest uses hover, which snaps to a nearby point)

### DIFF
--- a/src/hubbleds/components/dotplot_viewer/dotplot_viewer.py
+++ b/src/hubbleds/components/dotplot_viewer/dotplot_viewer.py
@@ -10,7 +10,8 @@ from hubbleds.viewers.hubble_dotplot import HubbleDotPlotView, HubbleDotPlotView
 @solara.component
 def DotplotViewer(gjapp, viewer: HubbleDotPlotViewer | None = None, data=None, component_id=None, title = None, height=400, on_click_callback = None):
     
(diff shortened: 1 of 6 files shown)


johnarban:
closing this as there will be a lot of changes needed

PR #434, opened 2024-06-26
Wire up piggybank mc total

patudom:
This relies on https://github.com/cosmicds/cosmicds/pull/298 and styles the chip for the piggy bank score total.

I'm 99% sure I haven't handled the updating of the `GLOBAL_STATE.piggybank_total` correctly, plus as I said in the above PR, whether this should live in the global, local, or component state may be changing anyway, so this is mostly about the cosmetic update. 

The piggybank total does increment when mc questions are answered on guidelines, but I think mc questions in slideshows are not set up with the right callbacks, so those are not incrementing the total score yet.

This builds on #432 so that should probably get merged first.


patudom:
I merged Nick's revamped state content to here, so this should be current. All it really does now is it restyles the piggybank chip to look how it looked before. It is not correctly wired up to show the updating points.

patudom:
This has been replaced by #440 .

PR #438, opened 2024-07-01
Fix mc score

johnarban:
This PR (built on top of #435 ) creates the data models, and callback functions required for the **multiple choice** and **free response** questions. 

There are generic data models (`generic_question_model.py`) that forms the basis for the `FreeResponse` and `MCScore` models and their containers. This is primarily for convenience in accessing and updating. 

In `state.py` I have added 4 functions `get_free_response`, `get_multiple_choice`, `mc_callback`, `fr_callback` to allow easy access and initialization of the questions. The current setup waits for the `initialization` event to create any new question entries. This can also be easily setup to create the entries immediately, when ever the associated `ScaffoldAlert` is called.  

The `free_response` is set up without the `Ref` to avoid the double-click issue with the next button. 

I have created a test stage to demonstrate how this can be implemented. 

## Problem
A problem that I am having is that the `LOCAL_STATE` is retreived from the DB *after*(?) the stage has already loaded, causing the scores and responses to be reset. 

The stage tries to initialize the questions: The appropriate callback checks to see if they are present, and if they are not, then it will create it.  This causes `LOCAL_STATE` to update (and I am unsure of the exact order). This is written to the DB, and then when the DB is finally read, it gets the reset values. And then those finally are what determines the state the user sees. 

I have turned **off** writing to the database in the test page, as a way to test my, and _with a user with an entered score on the DB_ refreshing will show those correct DB values, while turning on writing to the DB will then reset those values.


johnarban:
CLOSE in favor of https://github.com/nmearl/hubbleds/pull/6

PR #448, opened 2024-07-08
Use database data in stages 4 + 5

Carifio24:
This PR adds in the student/class/all data in stages 4 + 5, and adds them to our glue/Plotly viewers. This primarily focuses on the aforementioned data hookups as well as making any necessary changes to the existing layout to work with the new state structure. The data is loaded asynchronously in tasks, and then the relevant glue setup is done as the various data items are loaded. At points this got a bit complicated, so if anyone sees any issues please let me know.

I tried to guard against what I could, but there might be some issues if you go to one of these stages without data, or if your student isn't in a class. I set up class 215 to be an active class using the Hubble story, so if you add your student to that class then things should work okay. To help with this, I've also added a "Fill Data Points" button at the top of Stage 1, which fills the student's data with the "dummy data" that we've used in the past (it's always the same 5 points).

This requires https://github.com/cosmicds/cosmicds/pull/303 and `glue-plotly>=0.7.4`, which adds some necessary error handling that I noticed was missing while working on this.


src/hubbleds/pages/05-class-results-&-uncertainty/__init__.py
```diff
@@ -1,535 +1,709 @@
...
+        if not student_data.components:
+            student_data = empty_data_from_model_class(StudentMeasurement, label="My Data")
+        student_data = GLOBAL_STATE.value.add_or_update_data(student_data)
+        student_data_added.set(True)
+
+
+    if measurements_loaded.value:
+        _on_student_data_loaded(True)
+    else:
+        measurements_loaded.subscribe(_on_student_data_loaded)
```
nmearl, line 210: This is a guard in case the async loading of the measurements isn't completed by the time this page renders?

src/hubbleds/remote.py
```diff
@@ -62,6 +66,25 @@ def load_spectrum_data(
 
         return spec_data
 
+    def get_dummy_data(self):
```
nmearl, line 69:
```suggestion
    def get_dummy_data(self) -> List[StudentMeasurement]:
```

src/hubbleds/utils.py
```diff
@@ -126,3 +133,61 @@ def data_summary_for_component(data, component_id):
         summary[f"{percent}%"] = (bottom, top)
 
     return summary
+
+
+M = TypeVar("M", bound=BaseModel)
+def models_to_glue_data(items: List[M],
```
nmearl, line 139: I appreciate the generics, but is it possible to want to convert a list of inconsistent `BaseModel` subclasses to a glue `Data` object, or do we always expect them to be the same?

src/hubbleds/utils.py
```diff
@@ -126,3 +133,61 @@ def data_summary_for_component(data, component_id):
...
+    data_dict = {}
+    if items:
+        t = type(items[0])
+        ignore = ignore_components or []
+        for field in t.model_fields.keys():
+            if field not in ignore:
+                data_dict[field] = [getattr(m, field) for m in items]
+    if label:
+        data_dict["label"] = label
+    return Data(**data_dict)
```
nmearl, line 152: This seems a bit more lax that cosmicds/cosmicds#303, do we risk creating `Component` vs `CategoricalComponent` elements in the `Data` object?

src/hubbleds/utils.py
Carifio24: I was imagining that all of the items would be of the same type - glue wants every component to have the same shape, so I think we want every model instance to have the same structure here. Technically this setup still could have an issue where the first item is a subclass of `M` with additional model fields, but I'm not sure if there's a way around that.

src/hubbleds/pages/05-class-results-&-uncertainty/__init__.py
Carifio24: Yeah, that's the idea. The measurements will probably have been loaded in a previous step, but that won't be the case if this is the first page that the student is coming to.

src/hubbleds/utils.py
Carifio24: Good call - I'll update this to use the same logic as in the cosmicds PR (actually, I'll probably refactor the component-choosing logic into its own function that we can reuse here).

Carifio24:
I've updated this (and https://github.com/cosmicds/cosmicds/pull/303) based on @nmearl's comments and resolved the merge conflicts, so I think this should be good to go now.

nmearl (approved):
Looks good to me.

PR #456, opened 2024-07-10
Implement stage 3

johnarban:
This PR 
- implements the Stage 3 distance measurements
- Updates the dotplot to handle multiple datasets
- Allows for (but does not yet) adding lines to the dotplots
 - I am still unable to get an x-position, but the links are ok
- This makes some corresponding updates in stage 1


johnarban:
One error I get is when switching to stage 5


image

johnarban:
Current known issues are
- The dotplot selection layer does not report the actual `x` position of the click, just the middle of the plot. Right now it just picks a random x location. 
- for some reason the Dotplot tutorial doesn't show up, but turning the flag on in the state editor makes it pop up whenever you click next
- When adding data to the dotplot, the range needs to be reset, otherwise user data might be out of the shown range
- There is currently a lot of duplicated code for adding and managing the example galaxy data managements, which makes me think I am not doing it in the best way. In particular it get's added in `_glue_setup` in stages 1 & 2, and the `add_example_measurements_to_glue` and `add_link` functions are duplicated.

PR #460, opened 2024-07-10
Setup AWS codepipline auto deployment

nmearl:
This is to support the automatic deployment on pushes to the main branch of the hubbleds repo.


setup.cfg
```diff
@@ -68,6 +68,7 @@ install_requires =
     solara
     solara-enterprise
     traitlets
+    ipyfilechooser
```
Carifio24, line 71: This should be pulled in by our `glue-plotly` requirement, now that we've added the `jupyter` extras.

nmearl: Cool, removed!

PR #469, opened 2024-07-11
Stage 5 - layout update, wire up mc and fr guidelines

patudom:
This updates the layout to match the voila version more closely and wires up the MC & FR questions.

Note that we still have the FR issue where simply entering a valid response in the box does not trigger the gate to the next guideline to open. You have to click outside the entry box for it to register.

I've also commented out the writing of the Stage 5 component state to the db because it was triggering a lot of errors during testing depending on what guideline I was on. We'll need to do some more testing to track those all down.

Note that FR & MC questions will still be stored to the db because those are in the story state, not the stage state.

### DIFF
--- a/src/hubbleds/pages/05-class-results-&-uncertainty/__init__.py
+++ b/src/hubbleds/pages/05-class-results-&-uncertainty/__init__.py
@@ -331,8 +331,8 @@ def Page():
                     }                    
                 )
 
-                with rv.Col():
-                    ViewerLayout(viewer=viewers["layer"])
+            with rv.Col():
+                ViewerLayout(viewer=viewers["layer"])
 
     # --------------------- Row 2: SLIDER VERSION: OUR DATA HUBBLE VIEWER -----------------------
     if COMPONENT_STATE.value.current_step_between(Marker.cla_res1, Marker.con_int3):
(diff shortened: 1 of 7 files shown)


src/hubbleds/pages/05-class-results-&-uncertainty/__init__.py
```diff
@@ -711,14 +723,12 @@ def update_class_slider_subset(id, highlighted):
...
             event_fr_callback=lambda event: fr_callback(event=event, local_state=LOCAL_STATE),
             state_view={
-                "hint1_dialog": COMPONENT_STATE.value.age_calc_state.hint1_dialog,
-                "hint2_dialog": COMPONENT_STATE.value.age_calc_state.hint2_dialog,
-                "low_guess": get_free_response(LOCAL_STATE, "likely-low-age").get("response"),
-                "high_guess": get_free_response(LOCAL_STATE, "likely-high-age").get("response"),
-                "best_guess": get_free_response(LOCAL_STATE, "best-guess-age").get("response"),
+                "low_guess": get_free_response(LOCAL_STATE, "likely-low-age"),
+                "high_guess": get_free_response(LOCAL_STATE, "likely-high-age"),
+                "best_guess": get_free_response(LOCAL_STATE, "best-guess-age"),
```
Carifio24, line 728: I think we do need to get the response here, since on this guideline I'm seeing what's below

Screenshot from 2024-07-11 11-10-44

Carifio24 (approved):

PR #473, opened 2024-07-12
Wait for data loading and glue setup to complete before showing stage 5

Carifio24:
This PR updates stage 5 so that the layout doesn't display until all of the data loading and glue setup has already finished. This is accomplished by creating a `data_ready` computed that is only true once the corresponding `<some>_data_added` refs have all been set to true. Until this happens, the page content is just a loading spinner.

There's also a little bit of logic involving the all measurements/summaries data (the data external to the current student's class). This data only gets loaded in here to get passed to glue, so if the data is already present in the glue data collection, but not in the local state, we don't make another effort to load it in. This is one of the heavier-payload endpoints that we hit, so it's nice to try and do it only when necesssary.

### DIFF
--- a/src/hubbleds/pages/05-class-results-&-uncertainty/__init__.py
+++ b/src/hubbleds/pages/05-class-results-&-uncertainty/__init__.py
@@ -46,16 +46,16 @@ def Page():
 
     solara.lab.use_task(_load_component_state)
 
-    # async def _write_component_state():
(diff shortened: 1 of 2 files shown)


patudom:
This is helping a LOT, so merging it in now even though there are still errors cropping up - Jon will track those down separately.

PR #478, opened 2024-07-12
Log more info about errors

patudom:
I added the extra logging only for errors, not warnings or other statuses. Others should feel free to add more if that would be useful, or remove anything that seems not helpful.

(From looking in the cosmicds repo, I didn't see obvious places where extra logging would be helpful, but feel free to check there too).


patudom:
Merging this since it's so minor - feel free to add additional logging later if needed.

PR #482, opened 2024-07-15
Add student data and layer toggle to stage 4 viewer

Carifio24:
Currently the first viewer in stage 4 (the line draw viewer) uses the class's data, when it should use the student's data. This PR fixes that.


Carifio24:
Switching to draft as we're actually going to want both the student and the class data in this viewer.

Carifio24:
This PR now updates the viewer to include both the student and class data, and adds a layer toggle component for this viewer.

patudom (changes requested):
Thanks for this, @Carifio24! There seem to be a couple of funky interactions between the fit line, the drawn line, and the layer toggle.

- The line draw tool is only placing a dot instead of drawing the line to the dot.
- Unchecking the "Class" layer makes the label in the layer toggle go away. That label will need to stay put.
Screenshot 2024-07-16 at 4 51 02 PM

- The best fit line displays both lines even if only 1 layer is checked.
Screenshot 2024-07-16 at 4 51 20 PM

- Just to keep track of this, we should rename the `line_draw_viewer` in all the relevant places to indicate that it handles both line drawing and best fit lines.

Carifio24:
@patudom I just pushed a commit that seems to be working well on my end (see the video below). Let me know how it works for you.

https://github.com/user-attachments/assets/18f52775-1d89-492e-ad11-dd033003c354

PR #488, opened 2024-07-16
Styling updates

patudom:
This plus https://github.com/cosmicds/cosmicds/pull/311 is my attempt to bring the layout more inline with what we have in the voila version.

Note that I have changed the `error` color from pink to a lavender/dark-indigo, so a few spots where we are currently using the `error` color for team interface things will be hard to read. The reason is that the error notification in the free-response validator is tied to the error color, and I couldn't figure out a way to specify it differently. Since I believe this is the only place where this color gets shown to students, I decided to choose colors that would provide sufficient contrast in the context where they are being used.

There were some places where things didn't work as I expected (solara & vuetify seem to handle things slightly differently), so there are some classes & ways of specifying things that seem redundant - I decided to go with what works rather than what is elegant.

Note also that I updated "Hubble's Law" in the title bar to be capitalized, but if you are using an existing student, the app will load the uncapitalized (Hubble's law) title from the story state. This seems like one of those things that doesn't need to be in the db.


patudom:
Merging since this is just cosmetic and it will be easier to see where there are problems as we test other things.

PR #495, opened 2024-07-17
Update bins once all data has loaded

### LINKED ISSUE #490: Histogram viewer bins & x-axis limits need updating
The "My Class" histogram viewer bins are not on integers like they should be, and all histogram viewers need a bit of padding on the x-axis limits.
Screenshot 2024-07-16 at 3 47 46 PM

Carifio24:
This PR fixes #490. I think the best way to make sure that the bins are correct is to just explicitly update them once all the data has loaded, rather than trying to ensure that the listener gets fired during setup. This PR adds in that explicit call. Since the data should generally not (actually, should probably never) change during the stage, this should be all that we need - but just in case, I've retained the data change message listener update here as well.


patudom:
This works for me. Thanks!

Issue #445, opened 2024-07-08
Update spectrum button interactions


patudom:
When user zooms in, clicking the "let me measure" button should not reset the zoom.

patudom:
For reference, this is the spectrum viewer behavior from the voila version. (Looks like the rest wavelength H-alpha marker did go away when the wavelength zoom range changed, which is different than what I remembered, but the key thing is that when the zoom level changes, the ability to measure the observed wavelength needs to remain active). Let me know if you have questions.

https://github.com/cosmicds/hubbleds/assets/12750048/7a0e3366-14a0-4c15-9c0c-70ce234924e9

patudom:
Fixed in #496

PR #496, opened 2024-07-17
Implement proper spectrum viewer functionality

nmearl:
This issue addresses cosmicds/hubbleds#445 and updates the spectrum viewer to behave as was originally implemented in the voila version.

### DIFF
--- a/src/hubbleds/components/spectrum_viewer/plotly_figure.py
+++ b/src/hubbleds/components/spectrum_viewer/plotly_figure.py
@@ -1,0 +1,58 @@
+"""
+TODO: This needs to be removed. We only keep this here because plotly keeps
+  the configuration setting hidden in the `FigureWidget` instance, and we
+  don't have access to it from Solara natively.
+"""
+
+import solara
+from typing import Callable, Any
+
+
+@solara.component
+def FigurePlotly(
+    fig,
+    on_selection: Callable[[Any], None] = None,
+    on_deselect: Callable[[Any], None] = None,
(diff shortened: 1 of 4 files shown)


patudom:
@nmearl, thanks for this - it's looking great! I have a couple minor requests. 

- In the voila version, we locked down the zoom tool to only allow zooming in the x direction. If the user can specify both an x & y zoom, we think it's too easy for them to accidentally zoom to a range where the relevant part of the spectrum isn't visible. Can you please change the zoom to be x-only?

- The other issue is a consequence of not having a button that specifically turns the measuring tool "on" (which your prior implementation would have avoided). Right now, when the zoom tool is on, it's unclear whether the user intends to make a measurement or zoom in, and if the user is sloppy with their mouse motions, they might accidentally trigger the wrong option and get confused. I remember we had this same situation in the voila version, and we fixed it by automatically turning the zoom tool off after 1 zoom in. If they want to zoom again, they have to click the button again. Can you please update that?

And Nick, if you are strapped for time we can have someone else work on those minor changes too - just let us know.

This video shows both situations, in case it's helpful.

https://github.com/user-attachments/assets/bfc43233-2887-4776-b6f4-4669556240aa

nmearl:
Thanks for testing @patudom. I've pushed an update. Please let me know if this is more in-line with the intention. Note that by default, plotly restricts all tools to an axis. It is possible to allow e.g. panning on both axes, but zooming on one. Let me know if it's a preference.

patudom:
This looks great, Nick - thanks! Merging now.

PR #497, opened 2024-07-18
Add back stage 2 slideshow

patudom:
This PR puts Stage 2 back into the story.

The slideshow runs, but the Stage 2 component state is not being written to the db, so I must be missing something. @Carifio24 or @nmearl, would you mind taking a look at this when you get a chance, please?

### DIFF
--- a/src/hubbleds/pages/02-distance-introduction/__init__.py
+++ b/src/hubbleds/pages/02-distance-introduction/__init__.py
@@ -1,5 +1,6 @@
 
 import solara
+from solara.toestand import Ref
 
 from hubbleds.components import Stage2Slideshow
 from hubbleds.state import LOCAL_STATE, GLOBAL_STATE 
@@ -38,10 +39,19 @@ def Page():
     logger.info("Trying to write component state for stage 2.")
     solara.lab.use_task(_write_component_state, dependencies=[COMPONENT_STATE.value])
 
+    step = Ref(
+        COMPONENT_STATE.fields.distance_slideshow_state.step
+    )
+    max_step_completed = Ref(
+        COMPONENT_STATE.fields.distance_slideshow_state.max_step_completed
+    )
+    slideshow_finished = Ref(
+        COMPONENT_STATE.fields.distance_slideshow_state.complete
+    )
+
     Stage2Slideshow(
         step = COMPONENT_STATE.value.distance_slideshow_state.step,
         max_step_completed = COMPONENT_STATE.value.distance_slideshow_state.max_step_completed,
-        event_on_slideshow_finished=lambda _: COMPONENT_STATE.value.complete.set(True),
         length = 13,
         titles = [
             "1920's Astronomy",
@@ -61,5 +71,8 @@ def Page():
         interact_steps=[7,9],
         distance_const=DISTANCE_CONSTANT,
         image_location=f"{IMAGE_BASE_URL}/stage_two_intro",
+        event_set_step=step.set,
+        event_set_max_step_completed=max_step_completed.set,
+        event_slideshow_finished=lambda _: slideshow_finished.set(True),
         debug = LOCAL_STATE.value.debug_mode,
     )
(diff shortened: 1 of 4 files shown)


src/hubbleds/pages/02-distance-introduction/__init__.py
```diff
@@ -1,45 +1,65 @@
...
+
+        logger.info("Wrote component state for stage 2 to database.")
+
+    logger.info("Trying to write component state for stage 2.")
+    solara.lab.use_task(_write_component_state, dependencies=[COMPONENT_STATE.value])
+
+    Stage2Slideshow(
+        step = COMPONENT_STATE.value.distance_slideshow_state.step,
+        max_step_completed = COMPONENT_STATE.value.distance_slideshow_state.max_step_completed,
+        event_on_slideshow_finished=lambda _: COMPONENT_STATE.value.complete.set(True),
```
Carifio24, line 44: There needs to be an `event_on_slideshow_finished` argument added to the `Stage2Slideshow` function, because currently this isn't passed in to the component.

Once that's added, this should be something like `Ref(COMPONENT_STATE.value.distance_slideshow_state.complete).set(True)`. The component state doesn't have a `complete` field so this would give an error (currently it doesn't execute because of the event not being passed in to the component).

src/hubbleds/pages/02-distance-introduction/component_state.py
```diff
@@ -1,15 +1,14 @@
...
+class DistanceSlideshow(BaseModel):
+    step: int = 0
+    max_step_completed: int = 0
+    complete: bool = False
+
+class ComponentState(BaseState):
+    stage_id: str = "distance_introduction"
+    distance_slideshow_state: DistanceSlideshow = DistanceSlideshow()
+
+COMPONENT_STATE = solara.reactive(ComponentState())
```
nmearl, line 14:
```suggestion
from pydantic import BaseModel
from cosmicds.state import BaseState
from hubbleds.base_component_state import BaseComponentState

class DistanceSlideshow(BaseModel):
    step: int = 0
    max_step_completed: int = 0
    complete: bool = False

class ComponentState(BaseComponentState, BaseState):
    stage_id: str = "distance_introduction"
    distance_slideshow_state: DistanceSlideshow = DistanceSlideshow()

COMPONENT_STATE = solara.reactive(ComponentState())
```

Also, be sure to inherit from the `BaseComponentState` to get component step progressions.

src/hubbleds/pages/02-distance-introduction/__init__.py
nmearl: Just a quick note: will need to be `Ref(COMPONENT_STATE.fields.distance_slideshow_state.complete).set(True)` (note the `fields`).

Carifio24: Yes, good catch, it should be `fields` rather than`value`

patudom:
@nmearl, @Carifio24, I'm getting a long list of terminal errors that ends with this. I'm wondering if it's complaining because Stage 2 doesn't have any markers. It's just the slideshow. Any ideas of what to look into? I can post what I have if that would be helpful, but I can't check if it's correct until I figure out what's causing these errors.

```
  File "/Users/Pat/anaconda3/envs/common-solara/lib/python3.11/site-packages/pydantic/_internal/_generate_schema.py", line 405, in _unknown_type_schema
    raise PydanticSchemaGenerationError(
pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema for <enum 'BaseMarker'>. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.

If you got this error by calling handler(<some type>) within `__get_pydantic_core_schema__` then you likely need to call `handler.generate_schema(<some type>)` since we do not call `__get_pydantic_core_schema__` on `<some type>` otherwise to avoid infinite recursion.

For further information visit https://errors.pydantic.dev/2.7/u/schema-for-unknown-type
```

patudom:
Ok, I think I've gotten around the problem by putting in 1 marker. Making progress now.

patudom:
This is all working correctly for me now. @Carifio24, would you mind taking a look, please?

PR #502, opened 2024-07-18
Add score to piggybank_total as part of mc_callback

patudom:
This goes with https://github.com/cosmicds/cosmicds/pull/315.

Not sure if this is quite the right way to do this, but it seems to work.

I'm using the `GLOBAL_STATE` here because the `piggybank_total` is displayed in the `cosmicds layout.py.` Note, though, that the points total seems like it should be a story state thing (`LOCAL_STATE`) rather than app state. @nmearl, is there a way for the `cosmicds layout.py` to access the story state?

### DIFF
--- a/src/hubbleds/state.py
+++ b/src/hubbleds/state.py
@@ -216,6 +216,7 @@ def mc_callback(
     """
 
     mc_scoring = Ref(local_state.fields.mc_scoring)
+    piggybank_total = Ref(GLOBAL_STATE.fields.piggybank_total)
     new = mc_scoring.value.model_copy(deep=True)
     logger.info(f"MC Callback Event: {event[0]}")
     logger.info(f"Current mc_scoring: {new}")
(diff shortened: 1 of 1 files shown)


Carifio24:
I would agree that a student's score feels more like a story state item to me. If we do want to make it a story state item, I guess what we could do is pass in a story state instance (typed as `BaseLocalState`) to the base layout, through which we could handle story state-level properties

patudom:
@Carifio24, I think we decided we wouldn't worry about the story vs. app state location for the piggybank_total for now. If that's the case, can you please give this PR a review as is? Thanks!

Carifio24:
This works well for me up until a page refresh. I can see that the `piggybank_total` is stored correctly in the database, but after a refresh the score shows zero.

I think the issue isn't with anything here, but with the fact that we only load the local state [in the layout](https://github.com/cosmicds/hubbleds/blob/main/src/hubbleds/layout.py#L24). We haven't really noticed this so far as we haven't really been using the global state aside from the student/class info, which is explicitly loaded separately in the base CosmicDS layout.

patudom:
I see what you're saying. Would it be easy to add the loading of the global state, and if so, can you please set that up, @Carifio24?

Carifio24:
Sure, I'll add that on to this branch.

patudom:
Given the updates in #518 and cosmicds/cosmicds#318, this is now passing the piggybank_total from the story state to the cosmicds base layout.

Issue #503, opened 2024-07-19
Sometimes story state gets overwritten


patudom:
I noticed while wiring up slideshow mc questions that sometimes the story state gets overwritten. So far, I've noticed this happening while I'm in Stage 1, but I will make a note if I see it happen elsewhere.

@nmearl, do you have any ideas what is causing this and how to fix it? 

Here are some comments from @Carifio24 via Slack:
"so we set up the reading/writing of the story & app states in the layout
and it's not supposed to write until those states have been loaded from the database
we do that [here](https://github.com/cosmicds/hubbleds/blob/main/src/hubbleds/layout.py#L51). But I'm wondering if this is similar to the issue with the line draw viewer not showing up because of a task not running again"

As an example of this, when I was in Stage 4, my story state in the db looked like this:
```
{
  "app": {
    "drawer": true,
    "speech": {
      "rate": 0,
      "pitch": 0,
      "voice": "",
      "autoread": false
    },
    "student": {
      "id": 4864
    },
    "classroom": {
      "size": 17,
      "class_info": {
        "id": 215,
        "code": "solara_test_06_13_2024",
        "name": "Solara Test 06-13-2024",
        "test": false,
        "active": true,
        "created": "2024-06-13T15:57:17.000Z",
        "updated": null,
        "educator_id": 1,
        "asynchronous": false
      }
    },
    "update_db": true,
    "speed_menu": false,
    "allow_advancing": true,
    "piggybank_total": 0,
    "show_team_interface": false,
    "loading_status_message": ""
  },
  "story": {
    "title": "Hubble's Law",
    "story_id": "hubbles_law",
    "debug_mode": false,
    "mc_scoring": {
      "scores": {
        "pro-dat6": {
          "tag": "pro-dat6",
          "score": 10,
          "tries": 2,
          "choice": 1,
          "wrong_attempts": 0
        },
        "interpret-velocity": {
          "tag": "interpret-velocity",
          "score": null,
          "tries": 0,
          "choice": null,
          "wrong_attempts": 0
        }
      }
    },
    "calculations": {},
    "show_snackbar": false,
    "best_fit_slope": null,
    "free_responses": {
      "responses": {
        "shortcoming-1": {
          "tag": "shortcoming-1",
          "response": "",
          "initialized": true
        },
        "shortcoming-2": {
          "tag": "shortcoming-2",
          "response": "",
          "initialized": true
        },
        "prodata-free-4": {
          "tag": "prodata-free-4",
[...]

PR #505, opened 2024-07-19
Remove extra app & story state writers in stage 1

nmearl:
I never removed the app and story state writers in stage 1 once we moved them to the base layout class. This may be at least part of the issue @patudom is seeing in cosmicds/hubbleds#503.


patudom:
Yes, this seems to fix the issue. Thanks, @nmearl! If I see it happening again I will reopen the issue.

Issue #492, opened 2024-07-16
Stage 4 viewers sometimes don't appear on reload


patudom:
I'm not sure if this is a thing that could happen to a student (or if this is only happening to me because I randomly jump around), but I frequently end up with no viewer in Stage 4. Going back to the intro stage, refreshing the page, and then coming back to Stage 4 does not help - sometimes when I've tried, this the title bar for the viewer goes away too. The only way I've found to bring the viewers back is to reset the stage state, which would be problematic if it's possible for this to happen to students going through sequentially.

<img width="1251" alt="Screenshot 2024-07-16 at 4 42 03 PM" src="https://github.com/user-attachments/assets/abfa8212-de8a-4d65-a016-47deb733e888">

Carifio24:
Do you see any errors in the browser console when this happens? Since the graph isn't appearing, my guess would be that there's an issue with loading Plotly.

patudom:
```Uncaught (in promise) ReferenceError: Plotly is not defined
    at VueComponent.mounted (eval at <anonymous> (VueTemplateRenderer.js:351:21), <anonymous>:30:5)
mounted @ VM7451:30Understand this error
```

Carifio24:
So something is causing Plotly to not be loaded, which should happen in the layout via the `PlotlySupport` component (inside of the layout). This component adds a `plotlyPromise` to the window which we then await in the line draw viewer.

Is there any particular sequence of events that causes this to happen?

patudom:
Not that I've noticed. Usually I'm on a guideline where the viewer is visible, but when I refresh the page, the viewer is not there. If I notice any particular patterns going forward, I will note them here.

Carifio24:
Recap for posterity of a dev meeting discussion:

This can be partially resolved by moving the `PlotlySupport` component in the HubbleDS layout outside of the `use_memo` call. Doing this reveals another issue where the entire component won't load. This is due to (at least) one of the `student_plot_data` and `class_plot_data` being an empty array, causing the if condition that displays the component to evaluate false.

patudom:
I haven't seen this issue since the above update was merged, so I am closing this.

PR #506, opened 2024-07-19
Always mount Plotly & MathJax support components

Carifio24:
This PR gives a partial resolution to #492 by moving the `PlotlySupport` component outside of the `use_memo` in the layout. As the discussion there mentions, there is another part to that issue which involves tasks and loading in the data. That issue is being a bit difficult to resolve (mainly because it only happens every 1 in ~10 refreshes for me) so I figured we could merge this and at least make things better for now.

### DIFF
--- a/src/hubbleds/layout.py
+++ b/src/hubbleds/layout.py
@@ -11,13 +11,10 @@ logger = setup_logger("LAYOUT")
 
 @solara.component
 def Layout(children=[]):
-    def _mount_external():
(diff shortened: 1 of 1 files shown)


nmearl:
Can I also suggest you move the math jax support one out of the `use_memo` for now, too? It also suffers from the not-loading-on-refresh issue.

Carifio24:
I've added that in as well.

nmearl (approved):

PR #508, opened 2024-07-22
Wire up stage 1 slideshow multiple choice questions

patudom:
This wires up the multiple choice questions for the doppler_slideshow & reflect_velocity_slideshow in Stage 1. 

I also wired up the student's stored speed of light & calculated velocity which were previously showing up as 0 on the last slide of the doppler slideshow.


patudom:
Note: #506 seems to have fixed a lot of the Mathjax issues we were having in this stage, but there is a case on the last slideshow where Mathjax is not correctly rendering values that get entered within the slideshow itself.

I think the issue is that Mathjax goes through and typesets all the equations that appear in the entire vue file before the c & v values on the last slides are known. It then displays the equations with the updated values as un-typeset latex above the typeset versions with the old values. 

@Carifio24 or @nmearl, is there a way to force Mathjax to re-render before going to that last slide?

Screenshot 2024-07-20 at 10 56 13 AM

PR #513, opened 2024-07-22
Load global state in layout

Carifio24:
This PR is a companion to  https://github.com/cosmicds/cosmicds/pull/316. The idea here is to load in the global state after the student ID has already been loaded, and then load in the measurements once we've got the student's story state.

### DIFF
--- a/src/hubbleds/layout.py
+++ b/src/hubbleds/layout.py
@@ -49,7 +49,6 @@ def Layout(children=[]):
 
         solara.lab.use_task(_load_measurements, dependencies=[])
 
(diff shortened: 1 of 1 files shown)


nmearl (approved):
Both this and cosmicds/cosmicds#316 look fine to me, but can you expand a bit on what issue this resolves, for posterity?

Carifio24:
I linked to a comment in the other PR, but  the issue is that we currently aren't using the global state from the database. We only get the student/class info, but other stuff (like the MC total) just get their default values. This also adds a place to run code after the student info/global state has already been loaded from the DB.

Issue #515, opened 2024-07-23
Student galaxy measurements are not appearing in table anymore


patudom:
After merging in #513, my student's data table is not getting populated with my velocity & distance measurements (but they are still in the db). Student 4864 if that helps.

This is from Stage 4.
<img width="801" alt="Screenshot 2024-07-22 at 10 14 56 PM" src="https://github.com/user-attachments/assets/90fb9f98-eb2d-4e5a-bf89-cf1dd0d759bd">

patudom:
I just tried reloading and I couldn't even get to Stage 4. There were a gazillion logger messages that scrolled through my terminal and this appears over and over:
```
Traceback (most recent call last):
  File "/Users/Pat/anaconda3/envs/common-solara/lib/python3.11/site-packages/reacton/core.py", line 1707, in _render
    root_element = el.component.f(*el.args, **el.kwargs)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Pat/Documents/GitHub/hubbleds/src/hubbleds/pages/04-explore-data/__init__.py", line 57, in Page
    solara.lab.use_task(_load_component_state)
  File "/Users/Pat/anaconda3/envs/common-solara/lib/python3.11/site-packages/solara/tasks.py", line 846, in use_task
    return wrapper(f)
           ^^^^^^^^^^
  File "/Users/Pat/anaconda3/envs/common-solara/lib/python3.11/site-packages/solara/tasks.py", line 840, in wrapper
    raise task_instance.exception
  File "/Users/Pat/anaconda3/envs/common-solara/lib/python3.11/site-packages/solara/tasks.py", line 259, in runner
    self._last_value = value = await self.function(*args, **kwargs)
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Pat/Documents/GitHub/hubbleds/src/hubbleds/pages/04-explore-data/__init__.py", line 51, in _load_component_state
    LOCAL_API.get_stage_state(GLOBAL_STATE, LOCAL_STATE, COMPONENT_STATE)
  File "/Users/Pat/Documents/GitHub/cosmicds/src/cosmicds/remote.py", line 138, in get_stage_state
    component_state.set(component_state.value.__class__(**stage_json))
  File "/Users/Pat/anaconda3/envs/common-solara/lib/python3.11/site-packages/solara/toestand.py", line 358, in set
[...]

PR #517, opened 2024-07-23
Fix free-response and add back prodata viewers

johnarban:
This PR fixes the free-response questions, to remove the solara-side reactivity of those questions. 

It also makes sure data get's linked and adds back the prodate viewer in stage 6 and the csv files. I didn't add in the class data, as in trying to do so it was causing the app to repeatedly reload (likely me not putting something in the right place)


patudom:
Per Slack discussion, the issue you were having with the class data was the same issue I described in #515. Recent PRs should have fixed this problem, so we can add back the class data in a future PR.

patudom:
This is great, @johnarban! The free response entries work very smoothly now. 🌟

PR #518, opened 2024-07-23
Pass local state into base layout

### LINKED ISSUE #515: Student galaxy measurements are not appearing in table anymore

Carifio24:
This PR, together with https://github.com/cosmicds/cosmicds/pull/318, is intended to help resolve #515. This PR updates the interaction with the base layout. This removes the passing in of a callable to use after the student data has loaded and uses our previous loading mechanism (which seems to work fine), and also now passes in the story state so that we can grab its piggybank total in the layout.

### DIFF
--- a/src/hubbleds/layout.py
+++ b/src/hubbleds/layout.py
@@ -19,35 +19,31 @@ def Layout(children=[]):
     student_id = Ref(GLOBAL_STATE.fields.student.id)
     loaded_states = solara.use_reactive(False)
 
-    async def _load_measurements():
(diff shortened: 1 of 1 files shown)


patudom (approved):
This is working for me with no issues! Thanks, Jon.

PR #520, opened 2024-07-24
Refresh MathJax if necessary

### LINKED ISSUE #519: Stage 1 Doppler Slideshow MathJax Issue on last slide
#506 seems to have fixed a lot of the Mathjax issues we were having in this stage, but there is a case on the last slideshow where Mathjax is not correctly rendering values that get entered within the slideshow itself.

I think the issue is that Mathjax goes through and typesets all the equations that appear in the entire vue file before the c & v values on the last slides are known. It then displays the equations with the updated values as un-typeset latex above the typeset versions with the old values. 

@Carifio24 or @nmearl, is there a way to force Mathjax to re-render before going to that last slide?

Screenshot 2024-07-20 at 10 56 13 AM

_Originally posted by @patudom in https://github.com/cosmicds/hubbleds/issues/508#issuecomment-2243142616_

Carifio24:
This is an attempt to resolve #519. It's seemed to work for me so far, but some of these issues have been a bit flaky so I'd appreciate another look.

### DIFF
--- a/src/hubbleds/components/doppler_slideshow/SlideshowDopplerCalc5.vue
+++ b/src/hubbleds/components/doppler_slideshow/SlideshowDopplerCalc5.vue
@@ -50,8 +50,9 @@
         >
           <v-lazy>
             <v-card-text
-                v-intersect="typesetMathJax"
+                v-intersect="mathJaxObserver"
                 class="pt-8"
+            >
               <p>
                 Great! Now we'll continue the velocity calculation process in the sequence of this pop-up window to give
                 ourselves some more space to work.
@@ -161,7 +162,7 @@
         >
           <v-lazy>
             <v-card-text
-                v-intersect="typesetMathJax"
+                v-intersect="mathJaxObserver"
                 class="pt-8"
             >
               <v-card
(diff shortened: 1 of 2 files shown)


patudom:
Still not working for me. The slideshow behaves differently when you are going through for the first time in that stage vs. if you've already entered in the speed of light in a prior run on the slideshow. If you've done the slideshow once, your speed of light is stored in the db (even though it appears blank on the 2nd to last slide of the slideshow), and when you get to the last slide, MathJax already has your speed of light value and works correctly.

If you refresh your stage state and go through the slideshow for the first time, the original pass of MathJax still has 0 as the speed of light and I get variations of this. (Note if you refresh the stage state, you will have to remeasure the observed wavelength on marker `obs_wav1` otherwise the db will have that stored as 0.)

Screenshot 2024-07-25 at 10 29 15 AM
Screenshot 2024-07-25 at 10 31 37 AM

patudom (changes requested):
See above comment

Carifio24:
Okay, this is now working for me. I think what really made the difference was to not increment the step to the last value until we had updated the student's speed of light value.

PR #522, opened 2024-07-24
Stage 4 connections part 1

patudom:
This updates the Stage 4 Line Draw Viewer & the Layer Toggle. We're working towards making available and displaying specific layers and buttons at the right times. This also adds the best fit galaxy.

Note, I must have done something wonky with git because the first 5 commits here were already merged in #516. They seem to have different commit numbers here and I'm not sure how to merge this correctly. @Carifio24, can you please take a look?


patudom:
This is replaced by #523.

PR #523, opened 2024-07-25
Stage 4 connections part 1 (take 2)

patudom:
This replaces #522. I messed up my git history by cherrypicking commits onto a new branch without rebasing correctly, so I ended up with some duplicated commits that had different commit hashes that git thought were different. @Carifio24 kindly helped sort this all out.

This updates the Stage 4 Line Draw Viewer & the Layer Toggle. We're working towards making available and displaying specific layers and buttons at the right times. This also adds the best fit galaxy.

### DIFF
--- a/src/hubbleds/components/line_draw_viewer/LineDrawPlot.vue
+++ b/src/hubbleds/components/line_draw_viewer/LineDrawPlot.vue
@@ -1,7 +1,7 @@
 <script>
 export default {
   name: "LineDrawPlot",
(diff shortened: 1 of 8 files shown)


patudom:
There are still a few duplicated commits here - I will take a stab at recreating my commits manually onto main..
Screenshot 2024-07-24 at 8 52 32 PM

patudom:
This is replaced by #524.

Issue #458, opened 2024-07-10
Stage 3 issues 


patudom:
Current known issues are
- The dotplot selection layer does not report the actual `x` position of the click, just the middle of the plot. Right now it just picks a random x location. 
- for some reason the Dotplot tutorial doesn't show up, but turning the flag on in the state editor makes it pop up whenever you click next
- When adding data to the dotplot, the range needs to be reset, otherwise user data might be out of the shown range
- There is currently a lot of duplicated code for adding and managing the example galaxy data managements, which makes me think I am not doing it in the best way. In particular it get's added in `_glue_setup` in stages 1 & 2, and the `add_example_measurements_to_glue` and `add_link` functions are duplicated.

_Originally posted by @johnarban in https://github.com/cosmicds/hubbleds/issues/456#issuecomment-2220804902_

PR #525, opened 2024-07-25
Stage 3 and 6 updates

johnarban:
This PR addresses #458 with updates so the dotplots have the verticle line tool, and to allow it to broadcast the position and to recieve a line position. 

This also adds in the class data to stage 6. There is some issue with stage 5 not seeming to load data. Using the same async structure as in stage 5 would be preferrable, to decrease load time, but this is stable.


patudom:
@johnarban, this is looking good - thank you! I've noticed some updates that will be needed in stage 3, but I am going to merge this now so we have the nearly complete Stage 6 and will note the Stage 3 issues in a separate issue.

PR #529, opened 2024-07-25
Update stage 5 viewer colors

patudom:
This is cosmetic and just brings the stage 5 viewer titles and marker colors in line with what we used in the voila version


patudom:
moving these to a newer pr because it was conflicting with other updates

PR #548, opened 2024-07-29
Fix best fit galaxy

### LINKED ISSUE #541: Stage 4 - Track down why best fit galaxy does not always appear on best fit line.
Screenshot 2024-07-26 at 6 02 54 PM

patudom:
This fixes #541. I accidentally passed the wrong index to the vue file (and it was an unlucky coincidence that the student I was testing with had the same slope as the class). Just to be extra sure, I also rounded the distance and velocity values in the same was as in the backend.


Carifio24 (approved):
This looks good to me. I guess rounding the values will technically push the point slightly off of the best-fit line, but with the values that we generally have I expect that this will be imperceptible.

PR #549, opened 2024-07-30
Stage 6 Connections

### LINKED ISSUE #527: Stage 6 remaining issues
- [ ] Hide best fit line legends for most of the guidelines (so they don't give away the answers to the multiple choice questions) and turn them back on at the correct point.

patudom:
This PR fixes #527 and should complete Stage 6.
- Best fit age legend should appear at the right times now, so the mc answers are not given away.
- Layers appear/disappear based on markers.
- The comparison between our class age and the HST key project age is fixed now. I decided to set `allow_too_close_correct` to be true because I tested out some values that were really close.

Screenshot 2024-07-29 at 8 12 47 PM

### DIFF
--- a/src/hubbleds/pages/06-prodata/__init__.py
+++ b/src/hubbleds/pages/06-prodata/__init__.py
@@ -1,6 +1,7 @@
 import solara
 from solara.toestand import Ref
 import reacton.ipyvuetify as rv
+from numpy import where
 
 # cosmicds
 from cosmicds.components import (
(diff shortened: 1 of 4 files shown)


johnarban (approved):
So I think this is good to merge. The only issue I can see is that the `LayerToggle` does not deselect the layer when it is hidden by one of the `hide_*`/`show_*` methods. This is probably something that can be fixed in it's own PR in the `LayerToggle` component

PR #554, opened 2024-07-30
Add stage 4 waiting room

Carifio24:
This PR adds an implementation of the stage 4 waiting room. The behavior here differs slightly from the Voila version in that we now activate the advance button when enough students have completed their data, rather than the abrupt transition in the Voila version.

At the moment this si a bit difficult to test. I was able to test the class data polling functionality by continually using a dynamic checking condition:

```python
    async def keep_checking_class_data():
        count = 0
        enough_students_ready = Ref(LOCAL_STATE.fields.enough_students_ready)
        while not enough_students_ready.value:
            points = load_class_data()
            count += 1
            ready = len(points) >= 100 - 10 * count
            # Some logging statements for checking values
            logger.info(100 - 10 * count)
            logger.info(ready)
            logger.info(enough_students_ready.value)
            logger.info("======")
            if ready:
                enough_students_ready.set(True)
            await asyncio.sleep(2)  # This sets how often (in seconds) we re-check the class data
```

Also, if you then want to reset the `enough_students_ready` in your student's story state, you can run

```sql
UPDATE cosmicds_db.StoryStates 
SET story_state = JSON_SET(story_state, '$.story.enough_students_ready', false)
WHERE student_id = <your-student-id>;
```

### DIFF
--- a/src/hubbleds/components/__init__.py
+++ b/src/hubbleds/components/__init__.py
@@ -21,3 +21,4 @@ from .dotplot_tutorial_slideshow import DotplotTutorialSlideshow
 from .id_slider import IdSlider
 from .line_draw_viewer import LineDrawViewer
 from .plotly_layer_toggle import PlotlyLayerToggle
+from .stage_4_waiting_screen import Stage4WaitingScreen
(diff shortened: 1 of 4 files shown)


Carifio24:
Marking as a draft again while I make some updates to handle some items from #660.

PR #563, opened 2024-07-31
Update dotplot line label

### LINKED ISSUE #557: The line hover legend currently gives too much information. Can it report only the x value, rounded to the nearest integer?

Carifio24:
This PR attempts to resolve #557. The hover line label now looks like the following:
Screenshot 2024-07-30 at 8 04 17 PM


patudom:
Thanks, Jon! We're going to need to pass a units prop for the line label because not every dot plot is a velocity plot. 

Screenshot 2024-07-30 at 10 28 53 PM

Also, can a similar fix be applied to the spectrum line label?
Screenshot 2024-07-30 at 10 30 34 PM

Carifio24:
I've added in the unit option and resolved some merge conflicts. Still working on the spectrum viewer.

PR #567, opened 2024-07-31
Improve mathjax rendering in doppler dialogs

nmearl:
This adds in a few fallbacks to and attempts to trigger rendering only when needed. I had though that perhaps the duplicate rendering was due to subsequent pages in the stepper of the doppler slideshow being mounted by "hidden", and that causes repeated calls to the intersection observer. The improvements in this PR do seem to help for the most part, but unfortunately, on the last step of the slideshow, I can sometimes get it so that there's two equations: one that's unrendered (just the raw latex), but with the rendered version right under.

Unfortunately, I don't think I can fix it before tomorrow, but it should not significantly hinder the students.

### DIFF
--- a/src/hubbleds/components/doppler_slideshow/SlideshowDopplerCalc5.vue
+++ b/src/hubbleds/components/doppler_slideshow/SlideshowDopplerCalc5.vue
@@ -41,722 +41,748 @@
         <v-window-item :value="0"
                        class="no-transition"
         >
-          <v-card-text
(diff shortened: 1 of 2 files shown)


patudom:
Thanks for working on this, Nick! It's definitely behaving better for me than it was before, but I see what you're saying about the last slide still being problematic sometimes. It's not a breaking issue, so let's not worry about it for tonight.

PR #572, opened 2024-07-31
Put team interface content under if statements

### LINKED ISSUE #551: Turn off all the team interface components throughout Hubble
- [x] Check that we are actually setting the debug flag to true for local development and make it false on deployed version
- [x] Create an example of applying the global debug flag to a component (Jon)
- [x] Put all the debug stuff under an "if" statement based on the debug mode flag. (Pat)

patudom:
This resolves #551.

Note that `show_team_interface` is set to False in the cosmicds `GLOBAL_STATE`, so you will need to change the value if you need to see the team interface. BUT if I change it locally to True in my cosmicds/state.py file, the value keeps getting overwritten from the db back to False, so in the short term, you will likely need to edit your story state to get it to stick.  (I think this was the same thing that was originally happening with the piggybank total)?

I'll open an issue to set up a better long term fix for this.

### DIFF
--- a/src/hubbleds/pages/01-spectra-&-velocity/__init__.py
+++ b/src/hubbleds/pages/01-spectra-&-velocity/__init__.py
@@ -222,7 +222,8 @@ def Page():
             measurement.student_id = GLOBAL_STATE.value.student.id
         Ref(LOCAL_STATE.fields.measurements).set(dummy_measurements)
 
-    solara.Button(label="Fill data points", on_click=_fill_data_points)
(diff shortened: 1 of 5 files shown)


patudom:
Given there may still be some flaky behaviors, @johnarban and I decided to leave the state editor components visible to students so they can use them to get around. Do not merge this before today's class. (The buttons have been removed in a different PR)

patudom:
Prior attempt at #939

PR #587, opened 2024-08-22
Use vue for intro slideshow instead of solara

patudom:
This builds off of #586.

This is a short-term fix to avoid some of the wonky formatting issues in the solara version of the intro slideshow (where the funding acknowledgement footer covers critical navigational content).

I don't know how to pass the WWT viewers in this setup, so I've skipped for now all the slides that have WWT content.


patudom:
I don't know how this will behave with the component_state setup once this is actually writing to the db, so I decided to just let the vue file handle the step changes for now since we aren't keeping track of what anyone is doing in the demo version.

johnarban (approved):

PR #588, opened 2024-08-22
Shorten demo

patudom:
This builds on #587.

I was planning to shorten the DS for tomorrow's demo by commenting out markers to skip, but that causes a whole host of errors when markers we specify in the python file don't exist. I'll try going through in the morning to see if there is a non-destructive, not-overly-messy way of doing this; else we will leave as is.

Meanwhile, this does a couple minor things:
- end doppler calculation slideshow after a few slides in.
- I added a front-end button to randomly select 5 galaxies, but I didn't know how to set that up based on the voila button, so it currently doesn't do anything. If @Carifio24 has a chance to set up the backend function to do this in the morning, that'd be great; else we can delete the button.


johnarban:
It has the "Fill Data" button. Since nothing is sent to the server it shouldn't matter if there are duplicates. I would suggest just deleting the button. But I will 'fill data' button on stage 3 so peole don't have to bounce back

johnarban:
Alternatively, we can use `_fill_data` for your new button

patudom:
But if people use the "fill data" button, all the measurements will be complete in the table and the sequence where we have them make their own velocity and distance measurements won't make sense..

patudom:
Ok how about this:
- "Fill Galaxies" button gives the user the standard set of 5 galaxies but leaves wavelengths/velocities/ang_sizes/distances blank.
- After the example galaxy lambda measurement/velocity calc, we show them their filled 5 galaxy table and move to stage 3. (I guess this would mean replacing any galaxies they actually picked with the default 5 galaxies, but I'm guessing no one will notice).
- They do example theta measurement/distance calc, fill their table and move to stage 4.

PR #595, opened 2024-08-22
Shortcut buttons for stages 1 & 3

patudom:
Buttons behave as follows:
- Stages 1 & 3 both have a "fill everything in this stage & move on to next stage" button at the top.
- Stage 1 - while user is on marker sel_gal3 to choose 5 galaxies, shortcut button picks galaxies for you.
- Stage 1 - on marker rep_rem1, shortcut button can fill all lambdas for you.
- Stage 3 - on marker rem_gal1, shortcut button can fill all thetas for you.

Can someone go through and check that it makes sense?


Carifio24 (approved):

Carifio24:
These buttons all worked as intended for me. I think the names make it pretty clear what one should expect to happen as well.

PR #599, opened 2024-08-23
Bring over demo changes that we want to keep

Carifio24:
This PR brings over changes from the demo version of the app that we want to keep going forward. Commits chosen based on a Slack discussion with @patudom and @johnarban.

Additionally, this sets up the WWT viewers in the new intro setup, along with the buttons that move to the various galaxies.


patudom:
I'm having some issues with the dot plot section in Stage 3, but I will go ahead and merge this and we can sort that out later.

Issue #584, opened 2024-08-15
DataTable


johnarban:
Setting `selected` in the **DataTable**  does not select the galaxy. I tried matching the format of the `indexedItems` that it is based on and a few other things, but couldn't really get this to work. But, it should be selectable from solara. This is also important during initialization, since we don't save the selected galaxies to the state anymore.

PR #601, opened 2024-09-10
Allow specifying initial data table selection

Carifio24:
This PR is intended to solve #584. A few notes here:

* I tried passing in a `selected` list as a slice of the `items` list, but the selected items aren't the same objects that are in the table array JS-side, so that doesn't work. So I ended up making the Python-side prop be `selected_indices`, since then we don't have the same identity issues JS-side since we're looking at primitives
* I would've preferred to run the logic that updates `selected` JS-side in the component's mounted hook, but it seemed that the values coming from Solara hadn't been populated yet (I am not sure why). The solution that I used here is to use watchers on the props. I suspect there's no way to know which one will get populated first, so this just runs the logic in both watchers.

So as an example, if you wanted the data table to have the first item selected, you would pass in `selected_indices=[0]` to the `DataTable` component.

### DIFF
--- a/src/hubbleds/components/data_table/DataTable.vue
+++ b/src/hubbleds/components/data_table/DataTable.vue
@@ -80,12 +80,30 @@
 </style>
 <script setup>
 module.exports = {
+  data() {
+    return {
+      selected: [],
+    };
+  },
   computed: {
-    indexedItems () {
+    indexedItems() {
       return this.items.map((item, index) => ({
         id: item.galaxy.name,
         ...item
       }))
+    }
+  },
+  methods: {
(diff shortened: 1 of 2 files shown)


patudom:
@johnarban, I know we use this functionality in the voila version, but I don't remember exactly where. Can you please go ahead and try testing this at the appropriate place in the solara version of the story? Thanks!

johnarban (approved):

johnarban:
This works. I can add in the implementation for stage 1, where we need this. It is less crucial in stages 3 and 4. Should I do that here, or put it in a separate PR

PR #604, opened 2024-09-13
Sync Zoom and Liner markers on Dotplots & Spectrum Viewers

johnarban:
This PR syncs the axis limits and vertical line marks on the spectrum and dotplots. 

Stage 3 gives an example of how to activate and deactivate linking using a reactive (computed in this example). 

Common logic for linking two variables with some transformation between them is handled in a function `sync_reactives` in the `hubbleds` utils.

This does not adjust the story guidelines or apply the appropriate sequencing.


patudom:
I can't speak to the back end logistics, but on the user side, this is looking very much improved!! I noticed a few lingering wonky things:
- If you zoom in to a region of the dot plot, the spectrum view zooms to the correct corresponding range. If you click "reset" in the dot plot, that zooms the dot plot back out to the full range, but the spectrum graph zooms out to this funny partial range.  (If you do the same thing in reverse, i.e. starting with the spectrum graph, the dot plot does zoom back out correctly).
Screenshot 2024-09-18 at 6 56 39 PM


These next few issues can be seen in the bottom screenshot:
- We should use the same "reset" icon in both graphs. (In the dot plot, it's a double arrow; in the spectrum viewer it's a single arrow).
- The order of the icons should be the same in both graphs. I think what makes sense is put them in the order we present them to the students in the guidelines, then make the dotplot icon order match (where they are in common). 
- In the spectrum viewer, we should have a vertical line to mark the measuring tool (like we do in the dot plot viewer). I thought it used to be a vertical line, but now I'm getting a double-headed horizontal arrow, which does not show a clear x-value.

Screenshot 2024-09-18 at 6 59 14 PM

patudom:
Merging this since it is now mostly working. Will move any lingering tasks to a new issue.

PR #609, opened 2024-09-13
Fix issues with text detection in intro and stage 2 slideshows

Carifio24:
For me, this PR makes the text detection from the speech synthesis component behave properly in both the intro and stage 2 slideshows.


patudom:
This is working well. I noted a small issue in #613, but we can fix that later.

PR #611, opened 2024-09-17
Restore WWT nudges in intro slideshow

### LINKED ISSUE #603: Put back the nudges to move on for the WWT screens in the intro slideshow

Carifio24:
This PR fixes #603. I took the duration to wait before nudging (which is 2 minutes, in ms), from the voila version. All of this was handled Python-side in the old setup, but since we're using `setTimeout` client-side to do the timing, I just moved all of the logic into the Vue template as we only need to store a few different values.

Note that this is still Vue 2, so to update an array reactively we need to use `this.$set`.

### DIFF
--- a/src/hubbleds/components/intro_slideshow_vue/IntroSlideshow.vue
+++ b/src/hubbleds/components/intro_slideshow_vue/IntroSlideshow.vue
@@ -990,7 +990,7 @@ module.exports = {
   data() {
     return {
       target: '',
-      timerDuration: 120000,
+      timerDuration: [300000, 180000, 180000],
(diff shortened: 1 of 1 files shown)


patudom:
@johnarban, I adjusted the nudge values and wanted to see if you think what I put in was reasonable. The original value was 2 minutes for each slide, but using the audio reader on the default speed takes almost a full minute for some of the text, so I figured if someone were trying to read all the text before the explore, that would only give them ~a minute to explore. I made the first slide length longer for people who really get into it, but we can change it if you think that's too much time total for the intro.

PR #643, opened 2024-10-08
Fix example galaxy data

johnarban:
This PR fixes example galaxy measurements to get the correct first or second measurement data. The function still only returns either the first or second measurement, but it could be updated to include both and then subsets could be used to show one or the other.


patudom:
Also, I'm wondering if there is filtering turned on or whether there is still something a little off. There should be 40 measurements, and one of them should be at ~-60,000 km/s. That value is not showing up here, and I am only getting 39 measurements on this branch (when it's not throwing the above error).

This is from voila:
Screenshot 2024-10-09 at 2 51 16 PM

Issue #642, opened 2024-10-07
selection_tool_widget green galaxy markers are sometimes tiny


patudom:
Sometimes, the green galaxy markers are very small like this, and I don't know what's causing it:

<img width="1259" alt="Screenshot 2024-10-07 at 3 31 40 PM" src="https://github.com/user-attachments/assets/3835f876-b343-4f06-9e7e-93a478ba170c">

(They should look like this:
<img width="1232" alt="Screenshot 2024-10-07 at 3 33 09 PM" src="https://github.com/user-attachments/assets/e7abaf52-1e93-49f3-b636-f5643f9aa297">
)

patudom:
I've noticed when the green dots are tiny, they behave as if they have been scaled relative to the world instead of the screen. (They get bigger when you zoom in). I will try explicitly setting that to screen.
<img width="819" alt="Screenshot 2024-10-10 at 1 12 49 PM" src="https://github.com/user-attachments/assets/d4310d2d-a28d-4aa6-9390-c4e7932a1098">

patudom:
We think this has been fixed.

PR #653, opened 2024-10-11
Stage 1 sequencing updates

patudom:
This PR updates the Stage 1 sequencing and layout to better match what we had in the voila version. I also put back in a placeholder viewer for the 2nd measurement dotplot, but we are currently only reading in the 1st measurement seed data, so we will need to update that.

The main changes are:
- restore the instructions for galaxy selection that update on the guideline based on when galaxies are selected or the view is reset.
- update the dotplot sequencing. This does not match the voila version exactly, but it reflects various edits we've made to improve the flow. What I have here isn't final yet - we'll need to give it another pass once all the viewers are fully updated.
- A note about the dotplot sequencing - I changed the bin size from 75 to 74 to get the tall tower stack we had in the voila version (even though there we also used a bin size of 75) - I wonder if we had some half bin offset that was kicking half the points into the next bin).
- I added a suggestion from @Carifio24 to try to improve the situation described in #642. I'm seeing the problem less frequently now, but it hasn't gone away completely.
- I updated the layout to match the voila version, which looks neater to me than what we had before.


johnarban (approved):
This looks good to merge to me. 

Somethings I noticed that should be fixed in a later PR (that exist on main as well) are
 1) that the velocity for the real galaxies does not update if the student changes their measurement, they would have to press the fill velocities button. We either want to prevent changes, or automatically update them.
2) Selecting the galaxy does not seem to put it in the WWT view

Issue #652, opened 2024-10-10
Updating the position of the second measurement is no longer work


johnarban:
I only just noticed this on #651, but it seems to be present on main. This was definitely working before. It runs `add_example_measurements_to_glue()` when the measurement changes as it is supposed to (`update_second_example_measurement()` is poorly named and not responsible for this, it is only used to initialize the second measurement). 

So it runs, and the glue data set is updated, but the dotplot does not update. It does not seem to be an issue with @Carifio24 changes in #647, as it doesn't work on commits prior to it. I am unsure what is causing this.

johnarban:
Just a note to confirm that this is still broken in #657. The background glue object is being updated, but the dotplot does not show that.

Carifio24:
Sorry if I'm forgetting something obvious, but what is it that isn't updating? The dotplot seems to update for me when I change between the markers where we switch from first to second measurement.

https://github.com/user-attachments/assets/c7a86050-47c7-45a4-aca4-0e4351907987

johnarban:
Ah, no. So the issues that on `rem_vel1` clicking on the spectrum should make a new measurement, which should be reflected by the green dot on the dotplot. Instead the dot does not move. But when we advance to `dot_seq14` it suddenly shows up in the new location, if we got back to `rem_vel1` the dotplot still shows the old position. 

https://github.com/user-attachments/assets/07a6da1a-2fe2-48f3-9d77-540d47bec49a

PR #657, opened 2024-10-15
Set up dot plot viewer to display 2nd seed measurements

### LINKED ISSUE #655: Set up dot plot viewer to display 2nd seed measurements

johnarban:
This PR fixes #655 showing the second set of dotplot data. This enables this through several changes. 
- updates `create_dotplot_viewer` updates the arguments to be more declarative so that it is easier to customize each dotplot, allowing the user to set whether it shows the first, second, or full data sets for the example and seed measurements
- The example seed measurements are placed into 3 different glue data objects. This is needed to circumvent the dotplot's native method of handling the y-axis height when displaying a subset. We want the bounds to be set by the smaller data set (the first or second measurements) rather than the full dataset. If we use subsets, then the y-axis bounds are set by the larger (full) dataset, rather than just by what is visible.

This also fixes the spectrum wavelength label


Carifio24:
This should now fix #652.

patudom:
Per our conversation yesterday, the 2nd measurement is not showing up for me either, so I think we need to track down what's going on. It's really weird that there are things that work for @Carifio24 that don't work for me and @johnarban.

Carifio24:
Yeah, it's very strange that this works for me and not for both of you. For the purposes of comparison, my versions of our most relevant packages are:
```
glue-core==1.21.0
glue-jupyter==0.20.1
glue-plotly==0.10.0
pydantic==2.6.4
pydantic_core==2.16.3
solara==1.40.0
solara-enterprise==1.40.0
solara-server==1.40.0
solara-ui==1.40.0
```

patudom:
Mine are a bit different. Does this mean I set up my common solara environment incorrectly? 🤦‍♀️

These are the ones that are different. Should I try reverting to your versions and see if it works?  (It seems troubling to me that mine are newer and my version is one of the ones that isn't working..)

```
glue-jupyter==0.21.0  
pydantic==2.7.3
pydantic-core==2.18.4
```

patudom:
Just tried reverting to your packages and the 2nd measurement is still not appearing.

Carifio24:
So I was playing around with this some more, went back and re-did my student's first measurement, and then went forward again, and now the second measurement isn't updating with the line.

Carifio24:
Okay, after a long dive through the inner workings of the glue parts of all this, I think I finally figured out what the issue is. @patudom and @johnarban whenever you get a chance, let me know whether this works for you.

patudom:
It's working!! I will go ahead and merge. Thanks for tracking this down, @Carifio24!

Issue #616, opened 2024-09-20
Mathjax input boxes don't "remember" inputs


patudom:
- [ ] For example, in Stage 1, at marker `dop_cal4`, if you input the wavelengths into the equation box, open the calculation dialog, then close the dialog, in order to reopen the dialog, you have to re-input the wavelength values.

- [ ] Stage 3: est_dis3 marker always needs angular size input twice before advancing

Carifio24:
So in the Voila version, we used the fact that the app state was frontend-accessible to update the state on (I believe) a focus change.

I think in the Solara version we could achieve the same thing using an event callback that we invoke inside the input listener.

PR #663, opened 2024-10-16
Keep values present in dop_cal4 when correct

johnarban:
This PR allows for dop_cal4 to remember its value if it is entered and valid (issue #616)


patudom:
These are showing up for me as already filled in even though it's my first pass through this guideline.
Screenshot 2024-10-21 at 9 19 23 AM

johnarban:
I've updated this to use a separate variable on the `DopplerCalculation` state to keep track of whether or not to show the filled in value

johnarban:
@patudom this now has the appropriate handling on stage 3

PR #665, opened 2024-10-17
add demo label in css; don't retrieve db data

### LINKED ISSUE #659: Demo branch updates
- [x] Add clear labeling in layout when user is on the demo branch version of story (Pat)
- [x] Don't read in student's story or stage states or any measurements when in demo mode. (John)
- [ ] Reset server to make sure updates deploy (Jon)

johnarban:
Addresses first 2 items in #659  Adds a simple demo label using `::after` CSS psuedo-class, along with a companion cosmicds PR https://github.com/cosmicds/cosmicds/pull/345

image

### DIFF
--- a/src/hubbleds/assets/custom.css
+++ b/src/hubbleds/assets/custom.css
@@ -248,7 +248,7 @@ footer .v-card__text {
     DEMO VERSION TAG
     ================================
 *****/
-.app-title {
+header.cosmicds-appbar > div.v-toolbar__content > h3 {
     display: flex;
     flex-direction: row;
     align-items: center;
@@ -272,6 +272,8 @@ header.cosmicds-appbar > div.v-toolbar__content > h3::after {
     padding-block: 5px;
     border-radius: 10px;
     margin-left: 25px;  
+    
+    line-height: 1;
 }
 
 /***** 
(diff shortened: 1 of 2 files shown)


patudom:
The demo text was breaking awkwardly on smaller screens.
Screenshot 2024-10-18 at 12 16 41 PM

I updated as such. It's a bit cramped for 2 lines, but I think it's better than the awkward break. Ok?
Screenshot 2024-10-18 at 12 18 20 PM

(This needs a small update I pushed to https://github.com/cosmicds/cosmicds/pull/345).

patudom:
(The title bar defaults to 64px in height, but when I tried to use flex to make it expand beyond that, other bad things happened to the layout. If you know how to fix that, @johnarban, please feel free to make the text larger).

johnarban:
Your changes look fine to me. I added a commit to decrease the line-height. I think the strong red draws enough attention that we don't need to make the font too large.

Carifio24:
Looks good to me. My only comment is that some of the logger messages in the state (like e.g. https://github.com/cosmicds/hubbleds/blob/main/src/hubbleds/pages/01-spectra-%26-velocity/__init__.py#L98) are a bit confusing - the DB writing is correctly skipped and the logger message inside the API handler logs this, but you then see a "Wrote to the database" message.

This is minor, though. The functionality looks good to me!

PR #670, opened 2024-10-18
Prevent skipping stages

johnarban:
This PR fixes #577 in combination with https://github.com/cosmicds/cosmicds/pull/347. It stores the max stage index in the local state. This PR builds on top of #626

PR #683, opened 2024-10-30
Cherry pick to demo 2

patudom:
This cherry picks commits from more recent PRs to the demo branch.

### DIFF
--- a/src/hubbleds/assets/custom.css
+++ b/src/hubbleds/assets/custom.css
@@ -300,3 +300,10 @@ div.plotly .hoverlayer g.axistext {
 .hide-in-demo {
     display: none;
 }
+.fr-unit-custom {
+    font-size: 1.25em;
+    line-height: 0.5;
+    align-self: center;
+    font-weight: bold;
+    padding-left: 0;
+}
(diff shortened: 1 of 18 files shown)


patudom:
The commits to prevent skipping ahead using the left navigation menu weren't working on this branch - is that because this branch does not read from or write to the db? And is it ok to allow skipping ahead? (If you try to jump to Stage 5 without having measurements, the app will crash).

johnarban:
@patudom This does not seem to have any commits from #670. Was this just in local testing? Preventing skipping should work without updating the db, as the necessary info is stored in the state. 

So far as stage 5 failing, we should just find some way to get around the issue, can you copy the error here?

patudom:
> @patudom This does not seem to have any commits from #670. Was this just in local testing? Preventing skipping should work without updating the db, as the necessary info is stored in the state.

Yeah, I cherry picked them locally and they did not prevent moving forward.

johnarban (commented):
> > @patudom This does not seem to have any commits from #670. Was this just in local testing? Preventing skipping should work without updating the db, as the necessary info is stored in the state.
> 
> Yeah, I cherry picked them locally and they did not prevent moving forward. 

Alright, so the issues is actually in `cosmicds`. By default the `max_route_index` is None, which disables the effect. This should be defaulted to zero.

patudom:
This is working - thank you!

PR #686, opened 2024-11-01
Fix #679 by adding button to show doppler slideshow

### LINKED ISSUE #679: Allow student to reopen dop_cal5 slideshow if they want to
With #663, the dop_cal5 slideshow gets skipped if you already went through it correctly. We should set it up so students are able to reopen it to refer back to the content if they want to.

johnarban:
This PR fixes #679 adds a button, only visible during dop_cal4, and dop_cal5, for displaying the doppler calculation slideshow. It is only visible if `show_dop_cal4_values` has been set to true. The interface currently looks like

image

### DIFF
--- a/src/hubbleds/components/doppler_slideshow/SlideshowDopplerCalc5.vue
+++ b/src/hubbleds/components/doppler_slideshow/SlideshowDopplerCalc5.vue
@@ -3,6 +3,19 @@
       v-model="dialog"
       max-width="800px"
   >
+    <template v-slot:activator="{ on, attrs }">
+      <v-btn
+        v-if="show_button"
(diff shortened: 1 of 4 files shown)


patudom:
@johnarban, if I was on Marker `dop_cal5`, I was getting wonky behavior where the dialog would close when I clicked "next" in the slideshow instead of advancing to the next slide. I couldn't figure out why it was doing that, so I ended up taking a different approach. I treat the slideshow as belonging only to the Marker `dop_cal5`, which you can get to by clicking "next" from a correctly validated `dop_cal4` or `back` from 'che_mea1`. 

Can you let me know if this works for you, and if this behavior feels reasonable?

PR #693, opened 2024-12-09
Improve Question tagging and Track progress

johnarban:
This PR makes changes to help with accurately tracking progress and questions in the dashboard. 

- Make sure we know which stage questions belong to. This is done by passing the `COMPONENT_STATE` to the question creation functions (some additional logic is included to capture questions from students who have gone through the story already). 
- Calculate the progress in the stage. Because out stages can change, this makes sure the  `COMPONENT_STATE` tracks the progress itself via two new computed fields `max_step` and `progress` (a `float` on `(0, 1]`)


src/hubbleds/pages/01-spectra-&-velocity/component_state.py
```diff
@@ -114,6 +115,22 @@ class ComponentState(BaseComponentState, BaseState):
     velocity_reflection_state: VelocityReflection = VelocityReflection()
     reflection_complete: bool = False
     show_dop_cal4_values: bool = False
+
```
Carifio24, line 118: My one suggestion would be that we move these two computed fields into the `BaseComponentState` class so that we don't have to maintain this five times.

patudom:
@johnarban, I have a few additional comments:
- Are we sure the markers are indexed from 0? I am on the last marker on Stage 4, and it is giving me a progress of 106%, which makes me think they might be indexed from 1.
Screenshot 2025-01-02 at 2 23 00 PM

- Stage 2 is basically just the slideshow, so we might want to think in terms of measuring progress as the `slideshow_step / slideshow_length`
- Stage 6 is not padded with an extra marker, so I think `progress` should divide by `(self.total_steps)` instead of `(self.total_steps - 1)` 
- Per Jon's point about refactoring the computed fields to the common `BaseComponentState` class, I guess we could have a `marker_list_padded` prop or just pad Stage 6 so they are all padded.

src/hubbleds/pages/01-spectra-&-velocity/component_state.py
johnarban: Yeah, I thought this would be a good thing to have in the base component state as well. I will add that

johnarban:
closed because it now ahead of #774. See new PR #775

PR #699, opened 2024-12-11
Use SDSS 12 in selection and distance tools

Carifio24:
This PR updates the selection and distance tools to use the SDSS 12 dataset (which is now grabbing its deeper tiles from SkyServer). This PR requires https://github.com/nmearl/ipywwt/pull/5 and so will require a local reinstall of `ipywwt`.

In order to get this to work properly I had to slightly delay the setup of the foreground/background imagesets. Without this, using the SDSS 12 imageset wasn't working for me, and from adding some logging into the WWT JS I could tell that the problem was that the foreground/background update requests were happening before our default surveys WTML had finished being processed. The approach here is to put those calls into a `Timer`. I'm not a huge fan of this, but with the way that things are set up now I'm not sure that there's another option.

### DIFF
--- a/src/hubbleds/components/selection_tool/__init__.py
+++ b/src/hubbleds/components/selection_tool/__init__.py
@@ -1,15 +1,10 @@
-from random import randint
 
 import solara
-from solara import Reactive
-from hubbleds.base_component_state import BaseComponentState
-from glue.core import Data
 from reacton import ipyvuetify as rv
 from hubbleds.state import LOCAL_STATE
 from hubbleds.widgets import SelectionToolWidget
 from typing import Callable
-from hubbleds.state import GalaxyData, StudentMeasurement
-from solara.lab import Ref
+from hubbleds.state import GalaxyData
 
 
 @solara.component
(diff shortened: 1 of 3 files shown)


nmearl (approved):
This looks good to me. Is there a reason why the SDSS 12 dataset takes longer to load than the previous one?

Carifio24:
I'm not sure - it shouldn't take any extra time to register any particular imageset in WWT (it doesn't require fetching any of the data, just adding the `Imageset` object to the registry). Maybe because it's at the bottom of the WTML? I'm pretty sure it gets processed in order.

nmearl:
Hm. Maybe we should remove all the other image sets that we don't expect to use in order to speed up parsing.

patudom:
I agree with Nick that it would help to speed up the initial load if we can. Once it's loaded, I'm finding that most of the galaxies are resolving faster than with the older DR server, but sometimes they are still kind of pokey. As we select more and more of the galaxies during testing, will they get cached on the WWT server and be faster later?

PR #700, opened 2024-12-12
Add quick skip buttons and explain what is being skipped in Demo Version

### LINKED ISSUE #690: Demo version updates
- [x] ~~Don't lock down going to future stages~~
- [x] ~~set things up in a way that if you jump to a future stage, you can move through sensibly:~~
     - [x] ~~If you go to Stage 3, make sure you have galaxies selected with velocities measured~~
     - [x] ~~If you go to Stage 4, 5, or 6, make sure you have velocities & distances measured~~
     - [x] ~~If we use any of these autofills, pop up a snackbar that says, "Your data table has been auto-filled with the information needed to advance."~~
- [x] Add guidelines explaining what's being skipped
- [x] Easier plan - don't allow user to skip ahead using left navigation, but include clear buttons that allow user to quickly skip to next stage.

patudom:
This will close #690.

This PR does not allow user to skip ahead using left navigation buttons, but there are clear "jump ahead" buttons that will quickly take you to the next stage.

I've added guidelines explaining what is being skipped in Stages 1 & 3.

I've styled the sidebar so the selected stage is more readable.

### DIFF
--- a/src/hubbleds/assets/custom.css
+++ b/src/hubbleds/assets/custom.css
@@ -317,3 +317,9 @@ div.plotly .hoverlayer g.axistext {
     background-color: var(--success) !important;
     color: white !important;
 }
+
+/* Sidebar navigation active item */
+.v-list-item-group .v-list-item--active {
+    color: white;
+    background-color: var(--primary);
+}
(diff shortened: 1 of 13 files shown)


johnarban:
So there are a couple notes for this. 

- In Stage 5, it says the persons data is in orange, but it is not. Is this just a demo thing?
image

- Should we make the uncertainty tutorial skippable?

johnarban (commented):
see above comment

PR #712, opened 2024-12-13
Fix mistake in regression calculation in Plotly line draw viewer

Carifio24:
This PR fixes a mistake in the linear regression calculation in the Plotly line draw viewer - currently the slope when forced through the origin is calculated as sum(Y) / sum(X), when it should be sum(X*Y) / sum(X^2).


src/hubbleds/components/line_draw_viewer/LineDrawPlot.vue
patudom, line 1: Line 252 - should this be
`const b = yAvg - a * xAvg;`?

patudom (changes requested):

src/hubbleds/components/line_draw_viewer/LineDrawPlot.vue
Carifio24: Yeah, it should be that (this code path doesn't actually get used so this hasn't caused us any problems)

patudom (approved):

PR #718, opened 2024-12-14
Update limits on stage 5 viewers

### LINKED ISSUE #711: Stage 5 slider viewer limits are wonky
The slider viewer starts with a range that displays all the data, but it does not display the origin. If you hit the reset button, it includes the origin, but only shows the appropriate x-y range for the student being displayed, and then data points for other students fall off the graph when you slide between them.

What we want is for the graph to start with a range that is [0, max X for all students], [0, max Y for all students], and when the "reset" button is pressed, it returns to that starting range.

https://github.com/user-attachments/assets/516e7d10-11e3-4b58-91c1-d1a8374f39d2

Carifio24:
This PR fixes #711 by resetting the limits of each of the stage 5 viewers at the end of the glue setup.

### DIFF
--- a/src/hubbleds/pages/05-class-results-uncertainty/__init__.py
+++ b/src/hubbleds/pages/05-class-results-uncertainty/__init__.py
@@ -254,6 +254,9 @@ def Page():
         for viewer in (student_hist_viewer, all_student_hist_viewer, class_hist_viewer):
             viewer.figure.update_layout(hovermode="closest")
 
+        for viewer in viewers.values():
+            viewer.state.reset_limits(visible_only=True)
+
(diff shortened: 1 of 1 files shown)


patudom:
My axes for the viewer have disappeared with this commit. Do they appear for you?
Screenshot 2024-12-16 at 3 47 02 PM

Carifio24:
Oh weird - that also happens for me when I first load the page, but then is fine whenever I look at the graph after that:

https://github.com/user-attachments/assets/eb66adc6-67db-4d6e-8af2-abf611c98d2c

I'll investigate.

Carifio24:
@patudom I've just pushed a commit which fixes this for me. Does it work for you as well?

PR #719, opened 2024-12-16
Cherry pick spectrum toolbar updates to demo

patudom:
Bring #714 updates to demo.


johnarban (approved):
This seems good to merge for me. My only note is that there is a change to the `obs_wav2` guideline, and the demo does not show that guideline. This is still fine though

PR #729, opened 2024-12-20
Add 'my summary' subset to class ages histogram

### LINKED ISSUE #728: Add a layer toggle & subset for "my measurement" in class histogram
Screenshot 2024-12-20 at 10 25 21 AM

Carifio24:
This PR resolves #728 by adding a subset for "my age" to the "my class" histogram, as well as a layer toggle component for the class histogram viewer. The "my age" subset will be hidden when one of the percentage intervals is selected. This relies on https://github.com/cosmicds/cosmicds/pull/356.

@patudom I wasn't sure exactly which guidelines we wanted the layer toggle to be visible on, so feel free to adjust the logic that determines when the layer toggle displays. Also, let me know if I forgot to implement something that we wanted.

### DIFF
--- a/src/hubbleds/pages/05-class-results-uncertainty/__init__.py
+++ b/src/hubbleds/pages/05-class-results-uncertainty/__init__.py
@@ -14,14 +14,14 @@ from pathlib import Path
 import reacton.ipyvuetify as rv
 from typing import Dict, Iterable, Optional, Tuple
 
-from cosmicds.components import PercentageSelector, ScaffoldAlert, StateEditor, StatisticsSelector, ViewerLayout
+from cosmicds.components import LayerToggle, PercentageSelector, ScaffoldAlert, StateEditor, StatisticsSelector, ViewerLayout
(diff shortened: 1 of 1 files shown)


patudom:
@Carifio24, I updated the guidelines sequence and when the layers/LayerToggle appear. Can you please take a look and see if the flow makes sense and if I turned the layers on/off correctly?

One issue I've noticed is that if I move beyond this sequence and then back up, there is a rogue extra layer that appears in the layer toggle that doesn't seem to actually exist. I tried explicitly setting which layers I want the toggle to use, but the extra layer appears anyway. Is there a way to prevent this?

Screenshot 2025-01-02 at 4 21 21 PM

Carifio24:
@patudom The extra layer is the subset that the percentage selector creates in order to highlight part of the data - I've updated the layer toggle to ignore that layer.

Carifio24:
@patudom I forgot that my change to ignore the percentage selection layer also required a change to the layer toggle component itself - see https://github.com/cosmicds/cosmicds/pull/359.

PR #736, opened 2025-01-03
Add color constants

Carifio24:
This PR creates a `colors.py` file with a couple of color constants defined, and shows some examples (ID slider and stage 4) of how to use these in the story. This can be fleshed out by adding more colors and importing them as needed. The idea here is that we can then easily change the "color theme" of the story with only a few lines of code changes.

@patudom is this enough to get started with? Let me know if anything is unclear.


patudom:
This is my attempt at putting together a colorblind-friendly scheme across our Hubble viewers. Per @johnarban's suggestion, I focused on finding colors that work together within a single plot, rather than ones that are all uniquely distinguishable from each other across every viewer for all the different types of color-blindness.

If we move forward with these colors, we will need to be able to outline markers where colors have low-contrast against white, which @Carifio24 is working on. We should also vary the sizes of the scatter markers, but I haven't done that yet.

I updated all the guidelines I could track down where we had hard-coded a color description to use the variable names of the new colors, but it's possible/likely that I've missed some and we'll have to fix those as we come across them.

This is what the viewers look like for me.

Spectrum:
Screenshot 2025-01-06 at 7 43 25 PM

The dotplot dots are pretty small, so it is hard to tell the difference between my dot and the other dots. It would help if we could make all the dots bigger, and at least make my dot bigger.
Screenshot 2025-01-06 at 7 43 58 PM

My data vs. My class
Screenshot 2025-01-06 at 7 44 19 PM

My class vs. Other classes
Screenshot 2025-01-06 at 5 06 02 PM

My class vs Other classes vs. Other Students
Screenshot 2025-01-06 at 7 46 25 PM
Screenshot 2025-01-06 at 7 46 01 PM

Pro data
Screenshot 2025-01-06 at 7 42 13 PM

patudom:
@Carifio24, I couldn't tag you as a reviewer, I'm guessing because you started the PR, but can you please take a look and see if this looks ok? Thanks!

Carifio24:
> The dotplot dots are pretty small, so it is hard to tell the difference between my dot and the other dots. It would help if we could make all the dots bigger, and at least make my dot bigger.

I think the dot size here is primarily being constrained horizontally due to the high bin count - we have [74 bins](https://github.com/cosmicds/hubbleds/blob/main/src/hubbleds/pages/01-spectra-%26-velocity/__init__.py#L486) in this viewer. I could relax the spacing between dots a bit to help, though.

Carifio24:
@patudom If it helps for tracking down missed cases, I did some `grep`-ing and found three places Python-side (outside of `colors.py`) where we have a color literal rather than a defined constant. All three are in the selection tool widget:
[Line 122](https://github.com/Carifio24/hubbleds/blob/colors/src/hubbleds/widgets/selection_tool_widget/selection_tool_widget.py#L122)
[Line 128 (in commented-out code)](https://github.com/Carifio24/hubbleds/blob/colors/src/hubbleds/widgets/selection_tool_widget/selection_tool_widget.py#L128)
[Line 174](https://github.com/Carifio24/hubbleds/blob/colors/src/hubbleds/widgets/selection_tool_widget/selection_tool_widget.py#L174)

There are more instances in the Vue template files - it looks like mostly background colors and the color that we use in the distance tool canvas, but I admittedly didn't look through them all.

patudom:
> @patudom If it helps for tracking down missed cases, I did some `grep`-ing and found three places Python-side (outside of `colors.py`) where we have a color literal rather than a defined constant. All three are in the selection tool widget: [Line 122](https://github.com/Carifio24/hubbleds/blob/colors/src/hubbleds/widgets/selection_tool_widget/selection_tool_widget.py#L122) [Line 128 (in commented-out code)](https://github.com/Carifio24/hubbleds/blob/colors/src/hubbleds/widgets/selection_tool_widget/selection_tool_widget.py#L128) [Line 174](https://github.com/Carifio24/hubbleds/blob/colors/src/hubbleds/widgets/selection_tool_widget/selection_tool_widget.py#L174)
> 
> There are more instances in the Vue template files - it looks like mostly background colors and the color that we use in the distance tool canvas, but I admittedly didn't look through them all.

Thanks so much for checking for this, @Carifio24. I think it's ok to leave those alone because the WWT-viewer markers feel different to me from the scatter & histogram viewers.

Other than this, does this PR seem ok to you? Thanks!

PR #747, opened 2025-01-09
Bump glue-plotly version to 0.12.1

Carifio24:
This PR bumps our version of `glue-plotly` to 0.12.1 so that we can use the border styling added for scatter layers in that version. The relevant PR is [here](https://github.com/glue-viz/glue-plotly/pull/114), but what's important for us is a few updates to the layer state. In particular:

* `layer.state.border_visible` - set this to `True` to display borders
* `layer.state.border_size` - adjusts the size of the borders
* `layer.state.border_color` - sets the color of the border

For anyone who looks at the linked PR - the `border_color_match_layer` isn't something that we need - the intended use case there is for making the borders match the colors of a colormapped layer if you want to have unfilled circles.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -56,7 +56,7 @@ dependencies:
   - glfw==2.7.0
   - glue-core==1.21.0
   - glue-jupyter==0.21.0
-  - glue-plotly==0.10.1
+  - glue-plotly==0.12.0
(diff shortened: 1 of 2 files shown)


Carifio24:
I've bumped the version here up to 0.12.1 since we want the dotplot updates from that version as well.

PR #751, opened 2025-01-11
Cherry pick smaller WWT table to short demo

patudom:
(From #748).

This seemed like a pretty straightforward cherry-pick, but when I run it, I have to switch back and forth btw guidelines several times to get the green dots, so seems like this update is not solving the issue. ☹️ 

If someone else would like to give it a go, please let me know what you find. Don't forget that you have to set `CDS_DISABLE_DB=true` in the command line

In case it's in any way related, I got this error when I killed the process:
```
hubbleds Pat$ /Users/Pat/anaconda3/envs/common-solara/lib/python3.11/multiprocessing/resource_tracker.py:254: UserWarning: resource_tracker: There appear to be 2 leaked semaphore objects to clean up at shutdown

```

### DIFF
--- a/src/hubbleds/components/selection_tool/__init__.py
+++ b/src/hubbleds/components/selection_tool/__init__.py
@@ -23,9 +23,10 @@ def SelectionTool(
         def _add_widget():
             selection_tool_widget = SelectionToolWidget(
                 table_layer_data={
-                    k: [x.dict()[k] for x in LOCAL_STATE.value.galaxies]
-                    for k in LOCAL_STATE.value.galaxies[0].dict()
-                }
+                    k: [x.dict()[k] for x in LOCAL_STATE.value.galaxies.values()]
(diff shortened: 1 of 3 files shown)


patudom:
(Just to be clear about what's happening on this thread, I asked @Carifio24 to push his updates from #834 to this branch, where we knew the green dots were being wonky).

Results from my testing are that this is working well at Fast 4G or faster speeds. When I throttle to Slow 4G, the dots become unreliable again, and there seems to be a race condition. One time at Slow 4G  the green dots appeared on the very first guideline, like they didn't get the memo to have an opacity of 0. About a third of the time otherwise, the green dots did not appear on the 3rd guideline like they were supposed to. @Carifio24, did you throttle your network speed at all while testing?

Carifio24:
Ah no, I didn't throttle my network speed while testing this. Let me do that and try and pin down the race condition.

PR #758, opened 2025-01-21
Fix jiggly lines for demo

Carifio24:
Same as #757, but on the demo branch.

### DIFF
--- a/src/hubbleds/pages/05-class-results-uncertainty/__init__.py
+++ b/src/hubbleds/pages/05-class-results-uncertainty/__init__.py
@@ -311,6 +311,11 @@ def Page():
 
     logger.info("DATA IS READY")
     for name, viewer in viewers.items():
+        # We don't want to reset the class histogram's limits
+        # as we let its limits be controlled by the student histogram
+        # viewer's limits
+        if name == "class_hist":
+            continue
         visible_only = "slider" not in name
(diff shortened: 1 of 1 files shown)


patudom:
The lines are no longer jiggly! 🎉  But I'm now seeing some issues where the best fit line is not "on" when it is supposed to be (and I have to click the best fit line button twice to get them to display), and the limits on the viewers are wonky.

https://github.com/user-attachments/assets/fb1b6263-66e9-4118-9055-57740956d331

In Stage 5 on `cla_res1c`, I can't get the best fit lines to appear at all, and I get this terminal error: `invalid value encountered in divide`

The axes have no ticks other than 0,0.
Screenshot 2025-01-21 at 7 36 13 PM

patudom:
@Carifio24, I think this inadvertently broke something about the histogram limits. The all class and all student age histograms should look like this (with the limits of the all student age histogram): 
Screenshot 2025-01-23 at 10 30 33 AM

but they look like this on this branch (with the limits of the all class age histogram:
Screenshot 2025-01-23 at 10 30 26 AM

Carifio24:
@patudom I had forgotten that the class histogram viewer has the thing where we set its limits based on the student histogram viewer. My latest commit should fix the issue.

PR #760, opened 2025-01-21
Allow toggling DSS/SDSS

Carifio24:
This PR adds a button that allows toggling the background imageset between DSS and SDSS in the selection and distance tools. Whichever background the student has used will be the one that the widget attempts to set on each update trigger. This PR is built on #755.


patudom:
Is there a particular branch of ipywwt we need for this? This seems to be toggling between SDSS12 (which has the criss-crossing artifacts) and the old SDSS image set (which has a smaller footprint than DSS but does not has the criss-crosses).
https://github.com/user-attachments/assets/a0d550a0-61ad-41a9-b730-541592a5db9e

Carifio24:
@patudom I think my last commit should fix the issue with the wrong imageset being used.

PR #762, opened 2025-01-22
Hide footer on a short screen

### LINKED ISSUE #761: Don't let funding acknowledgment footer take up so much space on small screens
Hide the funding div if a small screen is detected.

Image

Carifio24:
This PR resolves #761 by hiding the footer on a short screen. The height is set [here](https://github.com/Carifio24/hubbleds/blob/6ac61a23558de72a62e476ce0056e7e2fd3e6d58/src/hubbleds/assets/custom.css#L206) if we want to adjust it.

Once we're happy with this, I'll do the same on the demo.

### DIFF
--- a/src/hubbleds/assets/custom.css
+++ b/src/hubbleds/assets/custom.css
@@ -202,6 +202,13 @@ footer .v-card__text {
     padding: 6px;
 }
 
+/* Hide the footer on a short screen */
+@media (max-height: 600px) {
+  footer.v-footer {
+    display: none;
+  }
+}
+
 .theme--dark .cosmicds-footer {
     background-color: var(--cosmicds-blue) !important; 
     color: #BDBDBD !important; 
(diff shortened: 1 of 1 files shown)


src/hubbleds/assets/custom.css
```diff
@@ -202,6 +202,13 @@ footer .v-card__text {
     padding: 6px;
 }
 
+/* Hide the footer on a short screen */
+@media (max-height: 600px) {
+  footer.v-footer {
+    display: none;
+  }
+}
+
```
johnarban, line 211:
```suggestion
/* also hide on narrow screens */
@media (max-width: 600px) {
  footer.v-footer {
    display: none;
  }
}
/* at 950px the lines wrap to be 3+ lines*/
@media (max-width: 950px) {
  .cosmicds-footer .v-card__text {
      line-height: 1;
  }
}
```

johnarban: We also need to hide in when the screen gets too narrow. To make it fit a bit better, I have adjust the `line-height` to 1 on medium-narrow screens

johnarban (changes requested):

PR #766, opened 2025-01-23
On demo, add button to choose 1 "random" galaxy on sel_gal2

johnarban:
add a button that allows users to select 5 random galaxies and add them to the measurements


patudom:
This allows the user to get past Stage 1, `sel_gal2` if they don't get the green dots as happens on some computers. The button says "random" but we are actually giving students the first of the dummy galaxy set, so if they want to use the shortcut "fill lambda" and "fill theta" buttons later, we will have the appropriate measurements to add to the table without having to swap out their galaxies.

PR #770, opened 2025-01-24
Fix issue with Doppler slideshow MathJax on a refresh

### LINKED ISSUE #735: Intermittent mathjax problem in doppler slideshow
With @Carifio24's fix in #520, the MathJax in the Doppler 5 Slideshow is mostly well-behaved now, but there is an intermittent problem that appears sometimes when you refresh the browser when the slideshow is open. The garbled mathjax appears, then if you move away and back to the garbled slide, it appears completely blank.

https://github.com/user-attachments/assets/7a86af3d-0793-427d-b833-a99c8761bb75

This doesn't always happen, though. Sometimes if you wait a second, the garbled mathjax typesets properly, and the slide behaves normally.

Hopefully this is an uncommon enough scenario that we don't have to worry too much about it?

Carifio24:
This PR delays displaying the Doppler dialog until the component state has loaded and makes a few Promise-based tweaks to the slideshow template that seem to fix #735 for me.

### DIFF
--- a/src/hubbleds/components/doppler_slideshow/SlideshowDopplerCalc5.vue
+++ b/src/hubbleds/components/doppler_slideshow/SlideshowDopplerCalc5.vue
@@ -954,8 +954,13 @@ mjx-mpadded {
 
 <script>
 export default {
+  mounted() {
+    this.refreshMathJax([this.$el]);
+  },
+
   methods: {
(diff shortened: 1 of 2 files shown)


patudom:
Closing this because the issue we intended to solve is fixed by #771.

PR #771, opened 2025-01-24
Exclude Doppler dialog flag from model dump

Carifio24:
This PR implements the idea discussed in https://github.com/cosmicds/hubbleds/issues/735#issuecomment-2612897972 - that is, if the student refreshes the page with the Doppler slideshow open in stage 1, they need to manually reopen it again after refresh.

For me, the app comes back to `dop_cal5` after a refresh, but it's indistinguishable from `dop_cal4`, and hitting next opens the dialog. Closing the dialog without completing it brings me back to `dop_cal4`, and completing the dialog brings me to `che_mea1` (the next marker), all of which I believe is the intended behavior.

### DIFF
--- a/src/hubbleds/components/doppler_slideshow/SlideshowDopplerCalc5.vue
+++ b/src/hubbleds/components/doppler_slideshow/SlideshowDopplerCalc5.vue
@@ -727,7 +727,7 @@
                 <div
                     class="font-weight-medium mt-3"
                 >
-                  Click <b>DONE</b> to close this pop-up window. The velocity will be filled in in the table for this
+                  Click <b>DONE</b> to close this pop-up window. The velocity will be entered in the table for this
                   galaxy.
                 </div>
               </v-card>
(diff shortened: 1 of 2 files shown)


johnarban:
For me, if I refresh, it brings me back to `dop_cal5` which has no next button. I would suggest having it either go back to `dop_cal4` if we load and the current marker is `dop_cal5`, or else change the `dop_cal5` guideline to show some other button.

Carifio24:
@johnarban With this PR, what do you see when you reload on `dop_cal5`? The `dop_cal4` dialog [should still show](https://github.com/cosmicds/hubbleds/blob/main/src/hubbleds/pages/01-spectra-%26-velocity/__init__.py#L708) (which is what I'm seeing on my end)

patudom:
I'm getting the same behavior as @johnarban, where the `dop_cal5` guideline does not have a "next" button. I think I can add a "next" button on `dop_cal5` that just opens the dialog.

patudom:
Ooh. I see now why in an earlier version of this, @johnarban set up a separate button to launch the slideshow on `dop_cal5`. The issue is that `dop_cal4` and `dop_cal5` display the same guideline. If you are on `dop_cal5` and the slideshow hasn't been completed, you don't meet the gate condition to advance to `che_mea1`, so the next button doesn't display. I had edited the sequencing to avoid John's button because it seemed confusing, but putting it back may be the easiest way forward. Will experiment a bit.
Screenshot 2025-02-03 at 6 56 17 PM

patudom:
We don't need the gate to `che_mea1` on the `dop_cal4/5` guideline because the slideshow is what controls advancing to `che_mea1`. Removing the gate puts back the next button which opens the slideshow. @Carifio24 can you please take a quick look to make sure this behaves sensibly and then merge? thanks!

PR #773, opened 2025-01-24
Use SDSS9 on main

Carifio24:
This PR updates the main app to use SDSS9, rather than SDSS12, in the WWT viewers.


johnarban:
This I am pretty sure can be merged. The switch to the HIPs is a good one

PR #774, opened 2025-01-27
Handle mc and fr questions as ref

johnarban:
This PR refactors the Multiple Choice and Free Response questions to be nested dictionaries that are updated using `Ref`s. This PR also makes sure the stage they belong to is a part of the question so they can be distinguished in the dashboard. 

The general sequence is 

```
free_responses = Ref(local_state.fields.free_responses).value.copy()['responses']
// do something ot the free responses
Ref(local_state.fields.free_responses).set({'responses': free_responses})
```

The `.copy()` seems necessary to prevent sharing modifications to `free_response` or 'mc_scoring` with other users. the mc and fr questions no longer have Pydantic models associated with them.


nmearl (changes requested):
This is looking good to me. But I haven't tried without @Carifio24's quick fix. Can you give this a shot after removing Jon's changes and see if you fix pre-filled FR or MC questions in two different app instances?

nmearl (approved):
Excellent -- looks good to me then!

patudom:
Great, thanks @johnarban & @nmearl. Do we still need https://github.com/cosmicds/cosmicds/pull/361 or do we want to undo those?

PR #798, opened 2025-02-05
Add bin highlighting feature to dotplot viewer

johnarban:
Introduce a bin highlighting tool and integrate it into the dotplot viewer, allowing users to toggle bin visibility for enhanced data analysis.

The BinHighlighter uses the `selection_layer` to get the current mouse position and display a histogram bin. This involves passing a message back to the server (which I think we wanted to prevent). 

The one issue remaining in it is - after zooming in, the bin highlight becomes pale (like it is placed behind a layer). if the mouse is moved quickly enough before the dotplot reorders the layers, it comes up alright.

image

In general, because of how small our bins are, I am not sure really how useful this tool would be

_I made a little solara app for testing it here: 
https://github.com/johnarban/test_highlight. In this repo is also an attempt at doing all of the highlighting client side (using javascript loaded in to the app). however, this would need to somehow be rerun everytime plotly rerendered, and I am not sure how to do that._


patudom:
I wrote the notes for #799 without reading your comment on this PR first, so I didn't realize that this works using messaging. I can say that @heywooddogwood and I both like the bin highlighter a lot, and I think we should be using fewer, wider bins, anyway, which would make the bin highlighter especially helpful. Also, we prefer the lighter gray highlight that you get when you're zoomed in.

But if there isn't an easy way to make this happen client-side, then maybe we will need to live without this functionality. Do @Carifio24 or @nmearl have ideas?

johnarban:
Yes, so https://github.com/johnarban/test_highlight/blob/master/test_highlight/components/PlotlyHighlighting.vue in that linked repo, starts to get highlighting working. I think it can work, if it can get rerun each time the plot is re-rendered. I have an idea, but we'll see

johnarban:
There will be an improved version of this. So closing for now

PR #805, opened 2025-02-07
use real student measurements

### LINKED ISSUE #811: Use actual student measurements now that we have them (rather than team seed measurements)

johnarban:
This PR will fix #811 switches us to use a locally stored json file as the source of example galaxy "seed" measurements. there is a corresponding csv file. I can share the SQL query required to create it. We can update and change the data to 

image

image

image


patudom:
Thanks for this, @johnarban. Can you please set this up so we can also pull a smaller random but consistent subset of the measurements to make the dots bigger?

PR #813, opened 2025-02-11
Demo: Make home tool for slider viewers not use only visible data

Carifio24:
The same as #812, but for the demo branch.


Carifio24:
@patudom @johnarban Were you both using https://github.com/cosmicds/cosmicds/pull/368? Just like #812, this depends on https://github.com/cosmicds/cosmicds/pull/368.

If I don't use that PR as well I see what you were seeing, but with it things look fine to me.

PR #819, opened 2025-02-18
Use local data in stage 5 class slider

### LINKED ISSUE #815: Demo class slider issue
I think this is analogous to the problem in #797, solved by #814.

I purposely made some wonky distance measurements that made "my class" data look like this.
Image

In the slider viewer, my wonky measurements (which you can see in the class-only viewer above) do not appear.
Image

Carifio24:
This PR resolves #815. As mentioned in that issue, the solution is similar to #814 - the only extra thing that we need to be careful about here is making sure to tag the local measurements with the correct class ID.


Carifio24:
Just pushed a commit that fixes that issue.

PR #822, opened 2025-02-20
Give demo user data that they need on a refresh

### LINKED ISSUE #789: Demo Branch: set up each stage so if you reload on a later stage, you have everything you need to proceed

Carifio24:
This PR resolves #789. I factored out some of the data-filling functionality so that it could be reused across multiple stages.

### DIFF
--- a/src/hubbleds/pages/03-distance-measurements/component_state.py
+++ b/src/hubbleds/pages/03-distance-measurements/component_state.py
@@ -73,8 +73,12 @@ class ComponentState(BaseComponentState, BaseState):
     
     @field_validator("current_step", mode="before")
     def convert_int_to_enum(cls, v: Any) -> Marker:
+        print(v, isinstance(v, int))
+        print(len(Marker))
         if isinstance(v, int):
+            print(f"Returning {Marker(v)}")
             return Marker(v)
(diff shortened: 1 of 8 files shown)


patudom:
If I refresh directly on Stage 4, the hubble viewer does not display, even though I have data in the table:
Screenshot 2025-02-27 at 11 08 03 AM

patudom:
(This is not happening consistently. Sometimes the viewer appears.)

src/hubbleds/pages/03-distance-measurements/component_state.py
patudom, line 1: Can these print statements be removed?

Carifio24: Oops, yeah, I'll get rid of those

Carifio24:
> (This is not happening consistently. Sometimes the viewer appears.)

Hmm, so after you press the "Graph" button to go `exp_dat1` the viewer just isn't there sometimes? I tried on my laptop and desktop but haven't been able to reproduce.

Do you have any console errors when this issue happens? This is one of the non-glue graphs, so I'm wondering if it has something to do with Plotly not having loaded yet.

PR #825, opened 2025-02-26
Address issue #799

johnarban:
Address issues raised in #799 - 
- reducing the number of bins
- always use full spectrum range
- Update tutorial data generation and slides
- Fix various issues related to marker comparisons and bounds resetting. This was originally mentioned in #821 and then later in #824. I wasn't aware of the #824 discussion and went with comparison operators.

Issue #821, opened 2025-02-18
Dotplot rerendering issues


johnarban:
In attempting to investigate  #816 and related issues involving the dotplot, I identified some issues with stage 1 and the dotplot. 
1) the dotplot component is re-rendered far more than necessary, with nearly every state change it seems to get rendered again. A possible fix would be to initialize a setup function as done in stages 4 and 5, via a `use_effect` with no dependencies. When one attempts to do this however, the dotplots render at the top of the page (rendering immediately when called, instead of where we finally place it)

Example:
```python
@solara.component
def Page():
...

def _setup_dotplots():
  v1 = create_dotplot_viewer(...)
  v2 = create_dotpolot_viewer(...)
  return {'plot1':v1, 'plot2':v2}
viewers = solara.use_memo(_setup_dotplots, dependencies=[])
...
with solara.Row():
  viewers['plot1']
...
```
Instead of `plot1` appearing in the `Row` it would appear above it, alongside `plot2`. `show=False` for the creation of the dotplot

2) the function to hide or ignore certain layers `hide_ignored_layers`, does not always get run when the dotplot is rerendered. Part of this is because the the `hide_layers` attribute does not change, however, once the dotplot rerenders, the `visible=False` seems to get reset. This can be made to run by placing refactoring  `hide_ignored_layers` to be a function outside of `_add_viewer`, then it will run, but still the `visible` attribute does not seem to stick. 

I attemped a few changes on this branch, like moving the `hide_ignored_layers` function, and getting rid of setting the zorder (which we shouldn't need).  https://github.com/johnarban/hubbleds/tree/johnarban/issue816

nmearl:
Okay. I haven't look at Stage 1 in awhile, but there are several problems going on here.

1. A lot of the computed functions are checking for `Marker` states using incorrect equality checks. That is, things like
   ```python
	 COMPONENT_STATE.value.current_step <= Marker.dot_seq4
	 ```
   should not work -- you need to check their actual values:
	```python
	COMPONENT_STATE.value.current_step.value <= Marker.dot_seq4.value
	```
   as we do in the built-in [methods on the `BaseMarker` class](https://github.com/cosmicds/hubbleds/blob/a266b4016e04c5ad97e2c71c9dfa2cfe2e8dee5e/src/hubbleds/base_marker.py#L37).

[...]

PR #826, opened 2025-02-26
Update dot plot to start fixing some issues

johnarban:
This PR builds on top of #825 addressing @nmearl's comments on #821. Update the dotplot to prevent some of the re-rendering, remove callbacks for manually setting the zorder, pass parameters as values to the dotplot. 

However see https://github.com/cosmicds/hubbleds/issues/821#issuecomment-2683621503

for the issue of the dotplot layers not staying in order or maintining visiblity.


patudom:
In the dot plot tutorial slideshow, if you zoom in on an x-range in the viewer, the slideshow closes unexpectedly. 

```
[2025-02-28 18:06:50][    INFO][ DOTPLOT][dotplot_viewer.py:326]:None: Zoomed
[2025-02-28 18:06:50][    INFO][ DOTPLOT][dotplot_viewer.py:332]:None: set x_bounds (6505.28, 20462.24)
[2025-02-28 18:06:50][    INFO][ DOTPLOT][dotplot_viewer.py:363]:None: Updating x_bounds
[2025-02-28 18:06:50][    INFO][ DOTPLOT][dotplot_viewer.py:102]:creating DotplotViewer: None
[2025-02-28 18:06:50][    INFO][ DOTPLOT][dotplot_viewer.py:326]:None: Zoomed
[2025-02-28 18:06:50][    INFO][ DOTPLOT][dotplot_viewer.py:335]:None: Bounds already set
```

patudom:
Other than the above hiccup with the slideshow, this branch is running well for me and seems quite smooth. I am not qualified to evaluate the handling of the solara reactive/effect/memo stuff, though, so it might be helpful to get another set of eyeballs on that aspect of this PR.

Carifio24:
So I haven't 100% worked this out yet, but I think the dialog is closing on a zoom because of the `x_bounds` reactive in the `DotplotViewer` being updated. For example, if I comment out [this line](https://github.com/cosmicds/hubbleds/blob/835ef9f0bd0370a52d5dc156ea9df3be92a877d9/src/hubbleds/components/dotplot_viewer/dotplot_viewer.py#L321) then the dialog won't close on a zoom. In particular, I _think_ that the problem is that this is causing the dotplot viewer to rerender, and then the slideshow to rerender, and AFAICT opening the dialog via the button doesn't actually update the `show_dotplot_tutorial_dialog` state variable (to see this, open the dialog and refresh, or open the dialog and look at the stage state in the database. Similarly, if you open the dialog via the state editor and then close it, the state variable still shows as true). Thus then the slideshow gets rerendered, it's closed.

I could be wrong on this, though - any thoughts are welcome.

Carifio24:
One thought that may or may not be helpful (as I don't know everywhere that we use the dotplot bounds) - instead of a reactive, maybe we could use callbacks on `x_min` and `x_max` of the viewer state?

Carifio24:
Yeah, I'm definitely thinking that this is the issue. We change the `dialog` v-model of the slideshow in the Vue template (i.e. JS-side) and so the Python side doesn't know that it's open if Solara re-renders it.

For an example of what I mean, try out the following mini-app:

`test_component.vue`
```vue
<template>
  <div>
    <p>{{ `JavaScript-side flag: ${flag}` }}</p>
    <v-btn @click="flag = !flag">Toggle flag (from JavaScript)</v-btn>
    <v-btn @click="toggle_flag">Toggle flag (from Python)</v-btn>
  </div>
</template>

<script>
export default {
  props: {
    flag: {
      type: Boolean,
      required: true,
    }
  }
}
</script>
```

`app.py`
```python
from typing import Callable
import solara


@solara.component_vue("test_component.vue", vuetify=False)
def TestComponent(flag: bool, event_toggle_flag: Callable):
    pass


@solara.component
def Page():
    flag = solara.use_reactive(False)

    TestComponent(flag=flag.value,
                  event_toggle_flag=lambda _: flag.set(not flag.value))

    solara.Text(f"Python-side flag: {flag.value}")
```

nmearl:
A simple fix for this issue is ensure you're setting the `show_dotplot_tutorial_dialog` in a callback instead of explicitly setting the `dialog` value in the vue component. See https://github.com/johnarban/hubbleds/pull/8.

PR #842, opened 2025-03-04
Dynamically set dotplot bin spacing and restore hover text

### LINKED ISSUE #841: Allow dotplot bin spacing to be set dynamically based on selected range
Goal is to make the bins small enough in width to avoid awkward translations between velocity & spectrum wavelength where the velocity appears not to line up with the observed wavelength of the line because the velocity value is not in the center of the bin.

johnarban:
Allow dynamic setting of bin spacing based on the selected range,

Restore hover text

Fixes #841

### DIFF
--- a/src/hubbleds/assets/custom.css
+++ b/src/hubbleds/assets/custom.css
@@ -281,7 +281,7 @@ footer .v-card__text {
 
 /* Hides the wavelength marker tracking the trace */
 div.plotly .hoverlayer g.hovertext {
-    display: none;
+    /* display: none; */
 }
 
  
--- a/src/hubbleds/components/dotplot_tutorial_slideshow/DotplotTutorialSlideshow.vue
+++ b/src/hubbleds/components/dotplot_tutorial_slideshow/DotplotTutorialSlideshow.vue
@@ -145,7 +145,7 @@
           class="black--text"
           color="accent"
           depressed
-          @click="step--"
+          @click="set_step(step-1)"
         >
           Back
         </v-btn>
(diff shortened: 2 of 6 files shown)


patudom:
@johnarban, the bin spacing looks great on all the relevant scales - thank you for working that out!

I fixed what seemed like a lingering bug on the dotplot slideshow.

There are a couple wonky things with the traces that we should clean up, but I'm not sure how to do that - can you please take a look?

- [ ] On the first dotplot guideline, there is a vertical trace at v=0 that shouldn't be there that persists until you click somewhere in the dotplot. (It appears in addition to the vertical measuring tool).  I think it's the trace that would appear in the dot plot when the user clicks to display the corresponding wavelength in the spectrum viewer, but we don't want it to appear until it's been clicked the first time, on marker dot_seq5.
Screenshot 2025-03-06 at 11 11 17 AM

- [ ] The spectrum viewer now has 2 labels - we only want the one on the bottom that displays the wavelength value. I'm guessing it appeared because we turned on the velocity label for the dotplot, but I'm thinking we should put the dotplot velocity value on the bottom for consistency anyway, and maybe if we do that, the 2nd label on the spectrum viewer will go away?
Screenshot 2025-03-06 at 10 44 39 AM

johnarban:
So - I have fixed the extra label on the spectrum viewer. 

For the label positioning, the `go.Scatter` layer and `go.Heatmap` layer used for the x data points, behave differently. For the heatmap, it places the label aligned with the cell (the center of the cell), the scatter trace allows you to place them on the axes depending on the options. 

If you think, well perhaps the SpectrumViewer could then go in the center, it cannot be placed directly. It can follow the spectrum, but only when the mouse is near the spectrum. 

So i don't think there is way to make the consistent.

patudom:
That's ok - thank you, John.

PR #844, opened 2025-03-06
Bring custom solara disconnect messages to main

### LINKED ISSUE #832: Bring solara template with custom error messaging to main
See demo PR #830

patudom:
This will resolve #832.


src/hubbleds/templates/index.html.j2
```diff
@@ -0,0 +1,725 @@
...
+                })();
+                if (!local) {
+                    const script = document.createElement("script");
+                    script.async = true;
+                    script.src = "https://www.googletagmanager.com/gtag/js?id=G-XXXXXXXXXX"; // Replace with actual tag ID
+                    document.head.appendChild(script);
+                    window.dataLayer = window.dataLayer || [];
+                    function gtag(){dataLayer.push(arguments);}
+                    gtag('js', new Date());
+                    gtag('config', "G-XXXXXXXXXX"); // Replace with actual tag ID
```
Carifio24, line 78: We need to put in the tag ID here, right?

Carifio24:
Mostly looks good to me - I just had one comment re: the Google Analytics tag.

Carifio24:
So @patudom and I were discussing that, rather than include the GA tag directly in the template, we should read in an environment variable instead. However, I'm unsure whether that will be possible - Jinja supports it, but we'd need to make sure that the appropriate variable(s) are given to the template environment.

It looks like the Jinja environment setup and rendering is done in Solara [here](https://github.com/widgetti/solara/blob/master/solara/server/server.py#L277). @nmearl, do you happen to know whether there's a built-in way in Solara to inject environment variables into the template?

nmearl:
It doesn't look like solara takes in any environment variables [during rendering](https://github.com/widgetti/solara/blob/8947be4d7ddbc891017e89d6a463a94f8ac0c355/solara/server/server.py#L390) of the template, unfortunately.

We'd probably have to submit a PR to solara itself to allow for passing in the python environment.

However, I'm curious where the desire to make the GA tag an environment variable comes from? Do we plan to have several different ones?

Carifio24:
@nmearl So in this case it wouldn't really offer an advantage, since this template is specific to the Hubble story. If we were going to reuse this template across stories, though, then I think we might want to have separate tags for each story

PR #858, opened 2025-03-10
Unify example data loading and fix tutorial data issues

### LINKED ISSUE #854: Duplicate data when going backwards from Stage 3 to Stage 1 (main)
```
Traceback (most recent call last):
  File "~/anaconda3/lib/python3.11/site-packages/reacton/core.py", line 1900, in _reconsolidate
    effect()
  File "~/anaconda3/lib/python3.11/site-packages/reacton/core.py", line 1131, in __call__
    self._cleanup = self.callable()
                    ^^^^^^^^^^^^^^^
  File "~/github/cosmicds_dev/hubbleds/src/hubbleds/components/dotplot_viewer/dotplot_viewer.py", line 161, in _add_viewer
    dotplot_view.state.x_att = viewer_data.id[component_id]
                               ~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "~/anaconda3/lib/python3.11/site-packages/glue/core/component_id.py", line 36, in __getitem__
    raise KeyError("ComponentID not found or not unique: %s"
KeyError: 'ComponentID not found or not unique: velocity_value'
```

### LINKED ISSUE #856: Stage 3 WWT doesn't always go to galaxy when selected in table
When you select a galaxy in the table, the action to go to the position of the galaxy in WWT only fires if the galaxy isn't the one that was already selected. That means that if you reload the page, the WWT viewer sets to the main SDSS fov, but the table thinks the galaxy is selected, so clicking it again does nothing. (The same problem happens when the user is invited to re-measure theta for the example galaxy because we've already been on that galaxy through the whole dot plot sequence).

johnarban:
Fix #854: Move example data loading into a single function to address missing tutorial data when transitioning backwards between stages 3 and 1. 

Fix #856: Don't set example galaxy on init. this forces it to be set and prevents it from changing which can prevent the distance tool from going to it again

Fix bunch of computed references in stage 3.

### DIFF
--- a/src/hubbleds/example_measurement_helpers.py
+++ b/src/hubbleds/example_measurement_helpers.py
@@ -18,6 +18,12 @@ from .utils import _add_link
 from .state import StudentMeasurement
 
 
+from solara import Reactive
+from hubbleds.remote import LOCAL_API
+from hubbleds.state import LocalState # used as a type
+import numpy as np
+
+
 
 def create_example_subsets(gjapp: JupyterApplication, data: Data):
     if EXAMPLE_GALAXY_MEASUREMENTS in gjapp.data_collection:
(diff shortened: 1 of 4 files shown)


patudom:
Thanks, @johnarban. This is mostly working pretty well. I think there is some funkiness with the bin sizes.

This is how it starts out, which looks correct.
Screenshot 2025-03-11 at 8 30 09 AM

But if I zoom in to a range and zoom back out, there are too few bins.
Screenshot 2025-03-11 at 8 30 28 AM

Also, do we intend for the measuring tool coordination to be only one way? (If you click in angular size, you get a marker in the distance dot plot, but not vice versa. The guideline makes it sound like this is what we intended, but it seems a little confusing for it to not behave the same way as the Stage 1 plots).

johnarban:
This should be fixed. But note that because of the inverse relationship, one may have smaller markers than the other when zoomed out. 
image

patudom:
Thanks, @johnarban. What is the thing that sets the angular size range? It basically goes almost to 0, which leaves a very long, empty tail on the right of the distance graph. Is it possible to restrict the angular size minimum value to be closer to where the data points start?
Screenshot 2025-03-11 at 5 28 59 PM

src/hubbleds/components/dotplot_viewer/dotplot_viewer.py
```diff
@@ -366,15 +366,16 @@ def on_zoom(bounds_old, bounds_new):
                 
             line_marker_at.subscribe(lambda new_val: _update_lines(value = new_val))
             vertical_line_visible.subscribe(lambda new_val: _update_lines())
-            def reset_hist_n_bin():
+            def reset_hist_n_bin(xmin = None, xmax = None):
+                print("reset_hist_n_bin")
                 if nbin_func is not None:
-                    dotplot_view.state.hist_n_bin = nbin_func(dotplot_view.state.x_min, dotplot_view.state.x_max)
+                    dotplot_view.state.hist_n_bin = nbin_func(xmin or dotplot_view.state.x_min, xmax or dotplot_view.state.x_max)
```
Carifio24, line 372: I don't know whether it would ever matter for us, but just a note that this could cause problems if either `xmin` or `xmax` is zero, as zero is falsy so `0 or <X>` is always `<X>`

src/hubbleds/components/dotplot_viewer/dotplot_viewer.py
```diff
@@ -366,15 +366,16 @@ def on_zoom(bounds_old, bounds_new):
                 
             line_marker_at.subscribe(lambda new_val: _update_lines(value = new_val))
             vertical_line_visible.subscribe(lambda new_val: _update_lines())
-            def reset_hist_n_bin():
+            def reset_hist_n_bin(xmin = None, xmax = None):
+                print("reset_hist_n_bin")
```
Carifio24, line 370: I guess this should use the logger here (or it's just a debugging statement that we want to remove).

Carifio24 (commented):
Looks good to me! I just had a couple of minor comments.

src/hubbleds/components/dotplot_viewer/dotplot_viewer.py
johnarban: This is a good point. I have resolved this be explicitly checking for None

src/hubbleds/components/dotplot_viewer/dotplot_viewer.py
johnarban: Removed that line. It was just for debugging.

Issue #840, opened 2025-03-04
Green dots much improved but still don't work with bad network speeds


patudom:
#751 & #834 now much improve the green dots situation on good internet speeds. If you throttle to slow 4G or worse, the green dots get flaky again. See if we can improve this further for slower internet.

patudom:
So far so good with #865.

Issue #447, opened 2024-07-08
Stage 1 - track down WWT race condition


patudom:
Sometimes the green dots that mark the available galaxy locations do not appear. They can be brought back by moving back a guideline and then forward again, but it would be nice if the student didn't have to do that.

patudom:
Now that John's [PR](https://github.com/nmearl/ipywwt/pull/4) has been merged, we think this should be fixed, but keep an eye out for this appearing again.

PR #865, opened 2025-03-12
Update with new iPyWWT widget

nmearl:
I've recreated the ipywwt widget to align more fully with the features of the original pywwt widget. In doing so, the `SelectionTool` has been refactored to be a full Solara component. Likewise, some changes have occurred for the `DistanceTool` component as well (though it remains ipywidget-based).

With the new ipywwt widget, there is a clear callback defined for when the wwt front-end is finished loading and ready to receive messages. This has allowed for implementing a loading spinner before rendering, as we as for removing some of the timers we had implemented to attempt awaiting the state of the wwt front-end.

We also now mount the wwt app in its own iframe, which requires having a separate process for hosting a local WWT instance. This instance will be used for any other widget which references the WWT research app.

This also seems to have resolved issues #840 and #447, but I encourage others to pull down this PR and give it a spin.


patudom:
Awesome!!! Thanks so much @nmearl. So far, I've consistently gotten the green dots on this PR, but it's hard to tell for sure if it's fixed because the issue is so intermittent. I think we'll have to trial this fix for a while before we know for sure that the problem has gone away for good.

I think WWT is missing a few things it needs - I'm getting DSS now instead of SDSS. @Carifio24, can you please make the relevant updates so we can merge this to the deployed demo? 
Screenshot 2025-03-13 at 10 37 31 AM

Thanks!

PR #869, opened 2025-03-13
Add example measurements to seed data

### LINKED ISSUE #859: When I remeasure example galaxy, dot counts should update to reflect the measurement change
Stage 1, rem_vel1 - when we display the student's measurement, if it has moved to a different tower, the tower heights should change (so we don't want to just overlay the new measurement on top of the bottom-most dot in the tower, which is what we think it's currently doing). And the old dot should go away, so we are not showing 2 orange dots at once.

Carifio24:
This PR is intended to resolve #859. As glue doesn't support "stacked" histograms (and the dotplot is just a reskinned histogram), we decided in a dev meeting earlier this way to work around this by including the student measurements in the example data to fake the same effect.

Thus the approach here is to update the relevant glue `Data` objects with the example measurements whenever they're updated. To do this, we only keep existing elements of the data that aren't from the current student, then add the new measurement(s) and update the existing `Data` object. Note that since we're potentially changing the data's size, we can't call `update_components`. So what we do here instead is to create a new `Data` object with the updated components and then call `update_values_from_data`.

### DIFF
--- a/src/hubbleds/pages/01-spectra-&-velocity/__init__.py
+++ b/src/hubbleds/pages/01-spectra-&-velocity/__init__.py
@@ -205,8 +205,6 @@ def Page():
     solara.use_memo(_glue_sync_setup, dependencies=[Ref(LOCAL_STATE.fields.measurements_loaded)])
 
     def _update_seed_data_with_examples(example_data):
-        print("Updating seed data")
-
         update_data = [
             [EXAMPLE_GALAXY_SEED_DATA, example_data],
             [EXAMPLE_GALAXY_SEED_DATA + "_first", [m for m in example_data if m.measurement_number == "first"]],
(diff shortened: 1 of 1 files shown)


patudom:
So my first measurement was in a tower of 5 (including me). If I update my measurement to a different velocity bin, I would expect the size of the tower I was originally in to drop to 4, but it stays at 5.

https://github.com/user-attachments/assets/8a8c3a8e-c6fe-484f-95d7-e2a5c9cdd541

Carifio24:
Just curious, if you re-measure to put it back into the tower of 5, does the tower height change then?

Carifio24:
Hmm...still haven't been able to reproduce that. Here's what it looks like for me

https://github.com/user-attachments/assets/78093fdd-e549-4639-97d3-b7c60080c0c2

patudom:
Weird. This shouldn't make a difference, but are you using a brand new student or an existing student? (I still get the behavior in my video either way). Is there a particular version of glue-plotly that I need to be using?

Carifio24:
For me, I've got:
```
glue-core==1.21.1
glue-jupyter==0.22.2
glue-plotly==0.12.1
plotly==5.24.1
solara==1.42.0
solara-enterprise==1.42.0
solara-server==1.42.0
solara-ui @ git+https://github.com/nmearl/solara.git@38651953335bc2b315cd5e538acfbd33e7dde468
```

Carifio24:
Okay, I've updated this based on our discussions earlier today - I think this is doing what we want now, but please let me know if that's not the case.

PR #870, opened 2025-03-13
Add gtag and update server error message

patudom:
Per our conversation today, this adds the google tag and updates the server error message so the user is not told to refresh the page when that will not help.


nmearl (approved):
Looks good to me!

PR #876, opened 2025-03-14
Use data for stage 3 bin limits

johnarban:
Use the data for the the stage 3 bin limits. I think we should also limit the max size of the galaxy. You can measure as high as 3600, which bunches all the measurements into a single super-tall and thin tower. Perhaps for the example galaxy we can limit the gaurd to 2 arcmin? Our largest galaxy i think is less than 3arcmin


patudom:
There is at least galaxy (id 296) in the sample that is legit nearby and large (~10') - just looked it up in WWT and it's M100.
Screenshot 2025-03-14 at 1 54 36 PM

Another (id 246) seems to be behind M106, so gets not unreasonably measured as ~25'.
Screenshot 2025-03-14 at 2 03 54 PM

So I'd say the guard probably wants to be at least 30' on the high end.

PR #877, opened 2025-03-15
Stage 1 and 3 updates

patudom:
Hodgepodge of Stage 1 & 3 updates - I think we are really getting close!
- fix a small bug in the slideshow
- manually filter example galaxy data to 40 points rather than relying on RNG which made it hard to balance using fewer points with showing all the wonky things we wanted students to see. For the record, I did this by filtering out the 2nd measurements, filtering out the class 209 as in the code, ordering the remaining velocities, then taking every 4th row and then putting enough back to get to 40.
- update Stage 3 sequencing and guidelines to reflect how the actual dotplot data look and to mimic how we end the dotplot sequence in Stage 1.


johnarban (approved):
On `dot_seq5`. Because the table updates, but the dotplots are not yet updated, perhaps we should hide them on that particular marker. I found myself checking them when measuring. The data selected looks good for both stage 1 & 3. 

Otherwise, this looks good to merge

PR #878, opened 2025-03-17
Bring updated WWT components to main story

Carifio24:
This PR brings the updated WWT components from #865 into the main story. These updates are largely the same as in that PR, although I did need to add in the functionality for responding to the selected measurement in the selection tool.


Carifio24:
@johnarban I agree re: changing the starting location - I often think that the WWT widget hasn't loaded since it just looks black. I'll let you and @patudom decide on where we want to start it. Whatever we do decide, I assume we'll want to do the same on the demo as well.

Looks like you fixed the CSS issue in #879 so I think this just needs the starting location update.

PR #883, opened 2025-03-19
Adjustments to WWT tools for slow internet

### LINKED ISSUE #882: WWT issues on slow internet
If a student is on slow internet and the student clicks "next" on the first guideline before the xml catalog has finished loading, they won't get SDSS as they should on the next guideline.

Possible solutions:

- [ ] don't let "next" button be active until WWT has finished loading.
- [ ] check if we still have the code that keeps trying to update from DSS to SDSS starting from the 2nd guideline on.

Carifio24:
This PR resolves #882 by making the following changes:
* Not allow the student to advance past the either of the first two guidelines in Stage 1 until the WWT widget has loaded
* As we were doing before, repeatedly try to change the background to SDSS at multiple points, in case this somehow doesn't happen the first time

This PR is on top of #878, as that PR makes significant changes to the WWT widgets that we want to be building off of here.

### DIFF
--- a/src/hubbleds/components/selection_tool/__init__.py
+++ b/src/hubbleds/components/selection_tool/__init__.py
@@ -24,6 +24,7 @@ def SelectionTool(
     galaxy_selected_callback: Callable,
     galaxy_added_callback: Callable,
     deselect_galaxy_callback: Callable,
+    background_counter: solara.Reactive[int],
     selected_galaxy: Optional[dict] = None,
     candidate_galaxy: Optional[dict] = None,
     on_wwt_ready: Optional[Callable] = None,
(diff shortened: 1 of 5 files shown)


patudom:
This only happened to me once and I can't seem to reproduce it, but the first time I ran this branch, the "next" button never appeared even after the WWT viewer loaded. If this happens to a student they will not be able to advance off the first guideline.

patudom:
I added a little "before-next" text.
It's kind of cramped with the long text on the "back" side of the guideline, but I think it's helpful to the user to see that there is a reason there is no "next" button yet. Does this look ok? This needs https://github.com/cosmicds/cosmicds/pull/373
Screenshot 2025-03-20 at 6 09 16 PM

PR #905, opened 2025-03-20
Fix #900 - Don't update mc/fr dictionaries directly

### LINKED ISSUE #900: Multiple choice & free response questions still overlap across users on deployed main

johnarban:
This PR fixes #900. The issue was that despite copying the free response and multiple choice dictionaries, they were still tied to their original objects. It seems likely that this was due to it being a _shallow copy_ which still refers back to the original object


nmearl (approved):
As @Carifio24 mentioned, we can remove the commented out stuff. It'll be preserved in the history.

PR #915, opened 2025-03-21
Update waiting room handling

patudom:
This updates the waiting room styling a bit and bypasses the waiting room if there are already enough students ready.

### DIFF
--- a/src/hubbleds/assets/custom.css
+++ b/src/hubbleds/assets/custom.css
@@ -317,6 +317,10 @@ div.plotly .hoverlayer g.hovertext {
     color: black !important;
 }
 
+.outline-warning {
+    border: 1px solid var(--warning) !important;
+}
+
 .student-warning.v-alert {
     background-color: #B20C28 !important;
     color: #FFF !important;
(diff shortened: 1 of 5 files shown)


Carifio24:
I'm running into a couple of issues here:
* The task cancellation [here](https://github.com/patudom/hubbleds/blob/update-waiting-room-handling/src/hubbleds/pages/04-explore-data/__init__.py#L219) is giving me an error when I first come to stage 4. I think something about the task way the task is being cancelled is problematic
* Even if I catch the error, it'll automatically move a student out of the waiting room because we set `enough_students_ready` in the task, which causes the line referenced above to be run

patudom:
Do I need to add something like this?
```
        try:
            class_ready_task.cancel()
        except RuntimeError:
            pass
```

Carifio24:
The exception that pops up isn't a `RuntimeError` but rather a `solara.tasks._CancelledErrorInOurTask`. But even if we catch it, there's still the issue where this auto-advances students through the waiting room rather than just activating the button.

PR #920, opened 2025-03-21
Fix #902 -  Nones being added to sample data

### LINKED ISSUE #902: Stage 1 dot plot is broken
It appears too early and the data tower is at the wrong value.

Image

johnarban:
By adding the example data directly to the sample data in `_update_seed_data_with_examples`, this circumvented the process @Carifio24 had put in place to take care of None values in the data. This fixes #902 by replacing `None` with `float('nan')` when adding it to data sample data.

### DIFF
--- a/src/hubbleds/pages/01-spectra-&-velocity/__init__.py
+++ b/src/hubbleds/pages/01-spectra-&-velocity/__init__.py
@@ -185,7 +185,7 @@ def Page():
                 )[1]
 
             for component in data.main_components:
-                update[component.label].append(getattr(measurement, component.label, None))
+                update[component.label].append(getattr(measurement, component.label, None) or float('nan'))
 
         new_data = Data(label=data.label, **update)
         data.update_values_from_data(new_data)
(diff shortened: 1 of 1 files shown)


patudom:
Confirming that this works for me, but we can let @Carifio24 have the final review on this.

src/hubbleds/pages/01-spectra-&-velocity/__init__.py
```diff
@@ -185,7 +185,7 @@ def _update_seed_data_with_examples(example_data):
                 )[1]
 
             for component in data.main_components:
-                update[component.label].append(getattr(measurement, component.label, None))
+                update[component.label].append(getattr(measurement, component.label, None) or float('nan'))
```
Carifio24, line 188: It may never be relevant here, but I think it makes sense to do something more like
```python
value = getattr(measurement, component.label, None)
if value is None:
    value = float('nan')
update[component.label].append(value)
```
If the value here is not `None` but is falsy (zero, empty string, etc.) the current line will change it to `nan`, which I suspect is not what we want.

Carifio24 (changes requested):

PR #930, opened 2025-03-24
Add "insurance policy" timeout to WWT gating

### LINKED ISSUE #928: Add "insurance" time out in case the WWT-ready gate does not open
I've encountered a few times a situation where WWT is ready but the gate fails to open. @Carifio24 suggested that we could implement a timeout that opens the gate if the wwt_ready signal fails.

Image

Carifio24:
This PR resolves #928 by implementing the timeout described there via tasks in stages 1 and 3. Currently the time is set to 10 seconds, @patudom feel free to adjust that.

I think the simplest way to test that is to comment out the `on_wwt_ready` arguments into the selection/distance tool components and check that the gate is opened after the specified amount of time.

### DIFF
--- a/src/hubbleds/pages/01-spectra-&-velocity/__init__.py
+++ b/src/hubbleds/pages/01-spectra-&-velocity/__init__.py
@@ -413,7 +413,7 @@ def Page():
 
     # Insurance policy
     async def _wwt_ready_timeout():
-        await asyncio.sleep(10)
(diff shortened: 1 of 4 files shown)


patudom:
Since the above comment is not easily reproducible and the stress test today showed that the timeout is needed, I am merging this and will keep track of the above in a new issue to keep an eye on.

Issue #898, opened 2025-03-20
Styling updates


patudom:
- [ ] #983
- [x] Give WWT viewer an outline to distinguish it from the background
- [x] Add tooltips to the top icons
- [x] hamburger icon for left menu does not appear
- [x] bottom banner is covering some content on page (like buttons that need to get clicked)
- [x] Give teal buttons on orange guidelines a white outline

johnarban:
further enhancements in https://github.com/cosmicds/cosmicds/pull/379

PR #931, opened 2025-03-26
Enhance slideshow and tutorial UI elements

johnarban:
Addresses several items from #898
- update slide show to use component state and save to database
- add outline to tutorial buttons in guideline
- add tooltips to toolbar and datatable buttons

### DIFF
--- a/src/hubbleds/assets/custom.css
+++ b/src/hubbleds/assets/custom.css
@@ -150,6 +150,15 @@ h4 {
     color:white !important;
     border-radius: 5px; 
     padding: 3px; 
+    padding-inline: 5px;
+    
+}
+
+.theme--dark .inline-guideline-button.tutorial {
+    outline: 1px solid black;
+}
+.theme--light .inline-guideline-button.tutorial {
+    outline: 1px solid white;
 }
 
 /* NOTE: .v-application based selectors will work, but often will need to be either !important or (also) have a higher specificity selector than the default Vuetify css. */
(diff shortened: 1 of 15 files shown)


patudom:
Lots of great improvements here - thank you, @johnarban!

There are a few problems I noticed and some places that could use the same improvements.

- [ ] The table buttons are not appearing for me in Stage 1 or Stage 3.
Screenshot 2025-03-26 at 11 37 06 AM

- [ ] In Stage 3, the distances are autofilling as I measure the angular sizes, and the part of the guideline that tells me to click the tape measure button never appears. (Don't know if this broke with this PR or if this was already misbehaving prior to this).

- [ ] In the intro, the progress calculator uses the guideline I'm currently on instead of the furthest guideline I've gotten to. If I made it to the last guideline, but then hopped back to an earlier guideline, I would expect my progress to stay at 100%, not revert to a lower % because I went backwards. (My recollection is that progress is always defined as the furthest point you've been to, not where you currently are, but we can discuss if I'm remembering that wrong). 

- [ ] I'm not sure why the outline styling is not being applied to the red button.
Screenshot 2025-03-26 at 11 34 08 AM

- [ ] Top buttons in layout (audio controls & light/dark theme toggle) need tooltips. (I know this will have to be in the cosmicds repo)

- [ ] Scatter viewer in Stage 4 buttons need tooltips.
Screenshot 2025-03-26 at 11 40 24 AM

PR #933, opened 2025-03-26
Modify some stage 4 handling to use computed values

### LINKED ISSUE #922: Stage 4 toolbar buttons not always disabled/enabled correctly on reload

Carifio24:
The goal of this PR is to resolve #922 by updating some of the logic in stage 4 for enabling/disabling various items to use computed values rather than rely on updates inside of a marker callback. See that issue for more discussion.


patudom:
This works well going forwards but the viewer markers all disappear if you move backwards from a certain point (and it turns out this failure mode existed before this branch too). 

I'm not adept enough at identifying what all the layers are to understand what's happening, but it feels like maybe we are inadvertently wiping both the dots and lines when we only want to make the lines go away, but they are coming and going together. I recall there were some issues with this in the voila version too and we needed to control independently whether the best fit line button is enabled/disabled (i.e. available to be pressed) and then whether the best fit line was being actively displayed or not.

https://github.com/user-attachments/assets/ada690f0-62cb-429c-b424-870e8e0a983b

Carifio24:
I was really confused about how the logic here was causing all of the layers to disappear - until I realized that there was a missing comma in the template that was causing Plotly to apply an update to all of the layers. Should be fixed now.

patudom:
Wow, pesky commas! Thank you for tracking that down. 

One more thing. Can you please make whether we show the age label with the best fit line also depend on the marker? Should be on starting with `age_uni4`
Screenshot 2025-03-27 at 11 09 00 AM

patudom:
Now line draw has a rogue legend. :(

Screenshot 2025-03-27 at 4 58 59 PM

patudom:
Also, if you reload, the age label is not reconstructed for some reason.
Screenshot 2025-03-27 at 5 07 21 PM

PR #937, opened 2025-03-27
Force foreground update as well in selection tool

Carifio24:
This PR implements what was discussed in the dev meeting today, which is that we should force an update of the selection tool foreground to SDSS as well on a marker change.


patudom:
Do we need to do the same in the distance tool? That's still giving me a lot of DSS.

PR #947, opened 2025-03-28
Layout updates

johnarban:
This PR, a companion to cosmicds [379](https://github.com/cosmicds/cosmicds/pull/379), fixes some of the css related to the footer, fixes a an issue with the step on the intro, and applies some minor simplifying changes to the intro structure


patudom:
Please add:

- [ ] Conditional to not display footer on small screen sizes
Screenshot 2025-04-01 at 5 35 54 PM



- [ ] Is there a way to tell the slideshow what height is available to it? Here, the next/back buttons should not be below the scroll since there is available space for them.
Screenshot 2025-04-01 at 5 40 13 PM

patudom:
Actually, now that I’ve moved off of this PR back to main, Nick’s [layout updates](https://github.com/cosmicds/cosmicds/pull/384) seem to be automatically doing both things that I mentioned above. I’m thinking the thing to do is to move the non-layout related changes from this PR to a new branch and close this. Would that be ok, @johnarban?

PR #948, opened 2025-03-28
Update foreground/background forcing in distance tool

Carifio24:
This PR updates the background/foreground behavior in the distance tool to try and fix issues where we don't get SDSS in Stage 3, as mentioned in https://github.com/cosmicds/hubbleds/pull/937#issuecomment-2759847526.


patudom:
#954 behaves well, and now that we have the viewers using SDSS + black sky background throughout Stages 1 & 3, I think this PR is no longer necessary. Can I close it @Carifio24?

Carifio24:
@patudom Yeah, if #954 solves the problem then we don't need this.

PR #964, opened 2025-04-02
Fix db load issues

nmearl:
This PR covers three main issues that we've been seeing:

## 1. Potential race conditions

There seems to be a problem wherein component states are not properly populated by the time their values are evaluated. This seems to be an issue that @patudom was experiencing, but I was unable to reproduce. Under the assumption that it's a race condition, I've converted all our `_load_component_states` to use `use_memo` instead of `use_task`. This means that the page should not load until the component state has been populated. I have also made the same change to the top-level `_load_global_local_states` function as well. **NOTE** Writing is still done asynchronously, so as not to cause any stuttering in the front-end user interface.

**BIG NOTE**: There is a weird caveat in the Solara docs about whether to include `dependencies=[]` in the hooks; from the `use_memo` docs:

> dependencies can take the value None, in which case dependencies are automatically obtained from nonlocal variables. If an empty list is passed as dependencies instead, the function is only executed once over the entire lifetime of the component.

What I interpret this to mean is that *if the intention is to run a function only once, or in response to a variable change **always set the `dependencies` argument***.

## 2. Further improvements to the routing

I ran across an issue in which the `push_to_route` utility function would not work appropriately when navigating to the root page of the app when the app is hosted at an endpoint. This is the same issue found in the navigation routing that was addressed in an earlier PR. Thus, I've included the appropriate fix to `push_to_route` function: it now must also be passed a `location` context.

## 3. Fully fix the wwt ready functionality

This was an annoying one (as I'm sure you all know!) but I finally realized something I had forgotten: `use_effect` is run _immediately_ **and then** every time its dependency changes. That means that upon page load, the `on_wwt_ready` callback passed to the `SelectionTool` was being run immediately, setting `wwt_ready` on the component state to true. This is why later attempts to set `wwt_ready` seemed to do nothing: since it was already `True`, no changes were being registered. To fix this, we simply check to see if we've actually received the ready notification from the WWT research app before we attempt to interact with the selection tool. Thus, I have also added back the step gates.

PR #966, opened 2025-04-02
Update Dotplot Handling

johnarban:
This PR updates the stage 1 dotplot handling to be more similar to the Stage 3 to be more explicit about when Dotplot-related stuff can be run. Instead of just looking at the state of the glue data object, it makes sure each step has been completed through `seed_data_setup` and `example_data_setup`. This also adds some better None value handling for the bounds that can be passed. 

It's quite possible that @nmearl's #964 makes this no longer an issue, since it forces the app to wait for the component state to be loaded (measurements are loaded in the `_load_global_local_states` he mentions). _I'd suggest testing if #964 fixes the issue, before merging this so we no what was causing the problem_

### DIFF
--- a/src/hubbleds/00-test-page/__init__.py
+++ b/src/hubbleds/00-test-page/__init__.py
@@ -51,7 +51,7 @@ def Page():
     
     solara.lab.use_task(_load_component_state)
     
-    async def _write_component_state():
+    def _write_component_state():
         if not loaded_component_state.value:
             return
 
--- a/src/hubbleds/components/dotplot_viewer/dotplot_viewer.py
+++ b/src/hubbleds/components/dotplot_viewer/dotplot_viewer.py
@@ -29,19 +29,28 @@ logger = setup_logger("DOTPLOT")
 from glue_jupyter import JupyterApplication
 
 
-def valid_two_element_array(arr: Union[None, list]):
-    try:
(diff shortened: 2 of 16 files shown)


patudom:
Thanks, @johnarban. It was while testing #964 that I saw the issue I reported on Slack, so I think we'll want this fix on top of #964. I was just thinking that since the Stage 3 dot plots behaved well with #964, that handling the Stage 1 dot plots how we handled them in Stage 3 would hopefully also fix the problem on #964. (I haven't been having issues with the Stage 1 dot plots prior to #964).

patudom:
Thanks, @johnarban. This seems to be back in the mode where my orange dot does not appear when I first get to this guideline, but it appears if I refresh. Are there any updates we can make to not need the refresh?

Screenshot 2025-04-02 at 4 54 53 PM
Screenshot 2025-04-02 at 4 55 05 PM

patudom:
Also on `rem_vel1`, remeasuring moves my dot to a new bin, but the moving dot is gray instead of orange.

https://github.com/user-attachments/assets/9e6ba2c6-b074-439a-bdfc-8ab79e109d8c

patudom:
This is working well for me now - thank you, @johnarban! We can merge it once we sort out what to do about the demo branch.

PR #973, opened 2025-04-03
Update demo for cosmicds 386

patudom:
This is a first pass attempt at applying the changes from #964 to the short demo branch so it will still work with https://github.com/cosmicds/cosmicds/pull/386.

The WWT selection & distance tools will fail because we had made a lot of changes to them on  only `main` over the last few weeks, so the versions of those components on the `short-demo` branch are behind. 

I'm posting this as a draft in case anyone else wants to take a look, but I can also come back to this in the morning.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -66,9 +66,8 @@ install_requires =
     pydantic
     python-dateutil
     reacton
-    solara==1.42.0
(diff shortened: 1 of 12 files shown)


patudom:
Superseded by #974

PR #988, opened 2025-04-04
Fix changes in stage 3 calling subscriptions in stage 1

### LINKED ISSUE #980: Stage 3: should remeasurement update take this long?
(See Slack development channel for video which was too large to post here)

johnarban:
This PR hopefully fixes #980. There were several issues.

1) With fbec07f122efa48caca41e2fb2f818a3604bfd03: Apparently modifying the elements via a ref to the indexed measurement was causing it to make the subscriptions in Stage 1 to fire. _I don't really know why...something, something, solara...something was a Ref to the same underlying object_
2) The biggest issue and a source of a lot of traffic was that our subscriptions were being recreated with each render. So they would run once for a change, and then twice, and then 3x, ........ Most of these were small, but they built up. At some point I knew this, and even guarded against it when using them.
3) We also had irrelevant calls to some functions, that were causing us to duplicate work
4) This also cleans up some of the repetitive setup in stage 1 and 3 and puts the functions in to a single place, and then they get called. I also made the setup much more symmetric, so that stage 1 & 3 should behave similarly. You will notice a difference between how `example_measurement` and `measurements` are handled in Stage 3 - I am not sure, but I think this is because their values are watched differently.

### DIFF
--- a/src/hubbleds/components/data_table/data_table.py
+++ b/src/hubbleds/components/data_table/data_table.py
@@ -4,7 +4,7 @@ from typing import Callable
 
 DEFAULT_HEADERS = [
     {
-        "text": "Galaxy Name",
+        "text": "Galaxy ID",
         "align": "start",
         "sortable": False,
         "value": "name",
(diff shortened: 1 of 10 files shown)


patudom:
The most recent commit is not fixing the issue for me. :(

patudom:
Nice work sorting this all out, @johnarban!! It is working flawlessly for me now. Thank you! 🙌

PR #999, opened 2025-04-09
Bump glue-plotly version to 0.12.3

Carifio24:
This PR bumps our glue-plotly version to 0.12.3 so that we get the latest updates to the dotplot sizing.


patudom:
Do we need to do this in the cosmicds repo too, since that's where the base `CDSPlotlyDotplotLayerArtist` lives?

Issue #885, opened 2025-03-19
Issue with Stage 1 shortcut button


patudom:
I encountered this while testing #878 with a brand new student.

I went through the first couple guidelines on Stage 1 and added a galaxy to my table. After verifying that worked, I used the Stage 1 -> Stage 2 shortcut button to advance so I could check the Stage 3 viewers. 

Later, when I came back to Stage 1, it looks like this. I think the action associated with the button didn't know what to do with the fact that I'd already added a galaxy to my sample.

<img width="1085" alt="Image" src="https://github.com/user-attachments/assets/cdac4230-b8d6-4ae8-bbec-96bcc24633bc" />

patudom:
Also, fix gate so if you have >5 measurements, you can still advance.

PR #1000, opened 2025-04-10
Adjust how shortcut buttons handle measurements

Carifio24:
While we've allowed advancing with > 5 galaxies now, part of the problem with #885 is the way that the shortcut buttons are handling measurements. I think the current handling method came from the demo, but caused problems in the main app since we're storing student measurements.

Basically, the current "fill" helpers clear out the current measurements locally and replace them with dummy measurements. This is fine in the demo, but causes problems in general. When the measurements list gets updated, these new measurements get added for the user in the database, but since we generally don't expect a student to "remove" a measurement, there's nothing that deletes the existing measurements in the database. This is ultimately what was causing the issue in #885.

The approach taken in this PR is to first fill in missing values in any existing measurements (up to the appropriate level dictated by the button location), and then add the necessary number of dummy measurements.


src/hubbleds/utils.py
```diff
@@ -128,12 +129,31 @@ def format_measured_angle(angle):
...
+def observed_wavelength_from_redshift(z: float, rest_wavelength: float) -> float:
+    return rest_wavelength * (1 + z)
+
+
+def rest_wavelength(galaxy: GalaxyData) -> float:
+    return MG_REST_LAMBDA if galaxy.element == "Mg-I" else H_ALPHA_REST_LAMBDA
+
+
+def distance_for_velocity(velocity: float) -> float:
+    return 13.8 * velocity
```
johnarban, line 142:
```suggestion
    return velocity / 70.85 # H0 = 70.85 km/s/Mpc for Age = 13.8 Gyr
```

patudom:
Don't know if this is a problem that existed before that I didn't notice, but the tables should display 0 decimal places.
Screenshot 2025-04-10 at 10 26 01 PM

patudom:
Also, we actually don't want to auto-calculate angular sizes and populate the db with this information. The `angular_size_for_velocity` calculation does not account for the large random variation in actual physical sizes of spiral galaxies, so this would lead to measured values that have too little noise. Stage 4 and beyond would be problematically dull.

The approach we took with the demo was to just replace whatever galaxies you'd already selected with the "default" set of galaxies in a bait & switch move like in a cooking show. I guess the problem is that you don't want to commit the default set of measurements to the db over and over again. I wonder if we should set up the button to also set the ignore flag in the db, so if someone uses the default measurements as a shortcut to progress through the story, their data automatically gets ignored?
Screenshot 2025-04-10 at 10 54 00 PM

Carifio24:
> I wonder if we should/could set up the button to also set the ignore flag in the db, so if someone uses the default measurements as a shortcut to progress through the story, their data automatically gets ignored?

I like this idea - we probably only want to populate the database with measurements that people are actually making. I can implement this if that's what we want to do.

patudom:
Yes, if this seems good to you, please go ahead and implement that. Thank you, Jon!

Carifio24:
@patudom I've updated the "fill measurement"-type methods here to now mark the current student as ignored. Note that this requires https://github.com/cosmicds/cosmicds/pull/394.

